PR #17, opened 2015-04-09
Added Neutral Resources Language Attribut 

Anheledir:
- To be used in an app package, portable libraries must define a NeutralResourcesLanguageAttribute on their main assembly (ie, the one containing code, not a satellite assembly). Set this to English.


nblumhardt:
Thanks very much!

PR #19, opened 2015-05-24
Added PurchaseProcessExample

fmondinelli:
The solution file is not included in the commit because I have VS12


nblumhardt:
Thank you for the example, it's a nice piece of code. I don't think we're going to add more examples to this repository at this time however (we have three in there currently). It would be great to see this published/blogged about somewhere though! :-)

PR #23, opened 2015-06-24
Make it safer not to configure a running statemachine by separating the configuration/operation into two interfaces

fraschfn:
- added two interfaces to the statemachine:
  - one for a not yet fully configured statemachine which only allows configuration stuff
  - one for a fully configurated statemachine which disallows configuration but allows operation stuff
- adjusted tests to work with the new interfaces

### DIFF
--- a/BugTrackerExample/Bug.cs
+++ b/BugTrackerExample/Bug.cs
@@ -12,7 +12,7 @@ namespace BugTrackerExample
         enum Trigger { Assign, Defer, Resolve, Close }
 
         State _state = State.Open;
-        StateMachine<State, Trigger> _machine;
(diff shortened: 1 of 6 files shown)


nblumhardt:
Thanks for the PR.

I've often thought I'd like to make this kind of improvement, though I feel like it's something that would be better shipped as part of a rethought v2. In the current version, I've been holding off making breaking changes to the API so far.

I will leave this open and think about how a new version of Stateless might be put together. Thanks again for the input!

nblumhardt:
Thanks again for the PR. I like the idea but don't think we'll make a change like this in the near future. All the best!

PR #26, opened 2015-07-06
Update project URL in Nuget package

BrainCrumbz:
Note: was looking also for an updated icon URL, but could not find any.


nblumhardt:
Thanks!

BrainCrumbz:
Cool. Please if you can, drop a line when package gets updated. Regards

PR #32, opened 2015-10-28
Fix typos in Enforce() calls in StateRepresentation.cs

jamesathey:
Very, very simple spelling corrections.


nblumhardt:
Thanks James!

PR #33, opened 2015-10-29
Add ToDotGraph() method

tlk:
Proposal for nblumhardt/stateless/issues/31

### DIFF
--- a/Stateless/DotGraph.cs
+++ b/Stateless/DotGraph.cs
@@ -1,0 +1,34 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+
+namespace Stateless
+{
+    public partial class StateMachine<TState, TTrigger>
+    {
+        /// <summary>
+        /// A string representation of the state machine in the DOT graph language.
+        /// </summary>
+        /// <returns>A description of all simple source states, triggers and destination states except those defined with PermitDynamic.</returns>
+        public string ToDotGraph()
+        {
+            List<string> transitions = new List<string>();
+
+            foreach (var stateCfg in _stateConfiguration) {
+                TState source = stateCfg.Key;
+                foreach (var behaviours in stateCfg.Value._triggerBehaviours) {
+                    foreach (TransitioningTriggerBehaviour behaviour in behaviours.Value.Where(b => b is TransitioningTriggerBehaviour)) {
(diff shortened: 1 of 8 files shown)


Stateless/StateRepresentation.cs
```diff
@@ -11,7 +11,7 @@ internal class StateRepresentation
         {
             readonly TState _state;
 
-            readonly IDictionary<TTrigger, ICollection<TriggerBehaviour>> _triggerBehaviours =
+            protected internal readonly IDictionary<TTrigger, ICollection<TriggerBehaviour>> _triggerBehaviours =
```
nblumhardt: Rather than making the fields `protected internal` could we wrap them in `internal` properties?

Stateless/DotGraph.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Linq;
+
+namespace Stateless
+{
+    public partial class StateMachine<TState, TTrigger>
+    {
+        /// <summary>
+        /// A string representation of the state machine in the DOT graph language.
+        /// </summary>
+        /// <returns>A description of all simple source states, triggers and destination states except those defined with PermitDynamic.</returns>
```
nblumhardt: Is it possible to show some evidence of the existence of dynamic triggers in the resulting graph, even though we can't determine the destination states? I.e. perhaps a pseudo-state called `?` that would be the target of dynamic triggers.. Any other options perhaps?

nblumhardt:
Looks great, just a couple of minor comments.

Is it also possible to include a README update in the PR that notes this method and links to some DOT visualisation options? Just a thought :-)

Cheers!

Stateless/DotGraph.cs
tlk: I have added code such that it will generate individual pseudo-states for each dynamic trigger. An alternative is to have each dynamic trigger point to the same pseudo state, but I think this will  make diagrams slightly less readable.

tlk:
Thanks for the feedback! Please let me know if there are any remaining/new issues or if you have more good suggestions! Also, I'll gladly squash this feature branch into a single commit before merge :-)

nblumhardt:
Thanks, looks great. Looks like I need to get CI for this set up :-)

tlk:
Thanks!

PR #38, opened 2015-11-26
DotGraph now renders OnEntry and OnExit actions with descriptions

cmendible:
Hi, I've been using stateless for several months now and one of my clients was missing meaningful descriptions for *Trigger Guards*, *OnEntry* and *OnExit* actions within the *DotGraph* feature.

For instance now it's possible to write the following code (which removes the compiler generated method names such as *<test>b__5_0* from the resulting string):

    machine.Configure(State.A)
        .OnEntry(() => { }, "Just entered A")
        .PermitIf(Trigger.X, State.B, () => true, "description");

And get the following *DotGraph* string:

    digraph {
     A -> B [label="X [description]"];
     node [shape=box];
     A -> "Just entered A" [label="On Entry" style=dotted];
    }

Hope you like it as much as our client  does and thanks for sharing this great code!!!
Carlos

### DIFF
--- a/Stateless/StateConfiguration.cs
+++ b/Stateless/StateConfiguration.cs
@@ -426,10 +426,14 @@ namespace Stateless
             /// trigger to be accepted.</param>
             /// <param name="guardDescription">Guard description</param>
             /// <returns>The reciever.</returns>
-            public StateConfiguration PermitDynamicIf(TTrigger trigger, Func<TState> destinationStateSelector, Func<bool> guard, string guardDescription = "")
+            public StateConfiguration PermitDynamicIf(TTrigger trigger, Func<TState> destinationStateSelector, Func<bool> guard, string guardDescription = null)
             {
                 Enforce.ArgumentNotNull(destinationStateSelector, nameof(destinationStateSelector));
-                return InternalPermitDynamicIf(trigger, args => destinationStateSelector(), guard, guardDescription);
+                return InternalPermitDynamicIf(
+                    trigger, 
+                    args => destinationStateSelector(), 
+                    guard,
+                    guardDescription != null ? guardDescription : guard?.Method.Name);
             }
 
             /// <summary>
@@ -444,7 +448,7 @@ namespace Stateless
             /// <param name="guardDescription">Guard description</param>
             /// <returns>The reciever.</returns>
             /// <typeparam name="TArg0">Type of the first trigger argument.</typeparam>
-            public StateConfiguration PermitDynamicIf<TArg0>(TriggerWithParameters<TArg0> trigger, Func<TArg0, TState> destinationStateSelector, Func<bool> guard, string guardDescription = "")
+            public StateConfiguration PermitDynamicIf<TArg0>(TriggerWithParameters<TArg0> trigger, Func<TArg0, TState> destinationStateSelector, Func<bool> guard, string guardDescription = null)
             {
                 Enforce.ArgumentNotNull(trigger, nameof(trigger));
                 Enforce.ArgumentNotNull(destinationStateSelector, nameof(destinationStateSelector));
@@ -453,7 +457,7 @@ namespace Stateless
                     args => destinationStateSelector(
                         ParameterConversion.Unpack<TArg0>(args, 0)),
                     guard,
-                    guardDescription);
+                    guardDescription != null ? guardDescription : guard?.Method.Name);
             }
 
             /// <summary>
@@ -469,7 +473,7 @@ namespace Stateless
             /// <returns>The reciever.</returns>
             /// <typeparam name="TArg0">Type of the first trigger argument.</typeparam>
             /// <typeparam name="TArg1">Type of the second trigger argument.</typeparam>
-            public StateConfiguration PermitDynamicIf<TArg0, TArg1>(TriggerWithParameters<TArg0, TArg1> trigger, Func<TArg0, TArg1, TState> destinationStateSelector, Func<bool> guard, string guardDescription = "")
+            public StateConfiguration PermitDynamicIf<TArg0, TArg1>(TriggerWithParameters<TArg0, TArg1> trigger, Func<TArg0, TArg1, TState> destinationStateSelector, Func<bool> guard, string guardDescription = null)
             {
                 Enforce.ArgumentNotNull(trigger, nameof(trigger));
                 Enforce.ArgumentNotNull(destinationStateSelector, nameof(destinationStateSelector));
@@ -479,7 +483,7 @@ namespace Stateless
                         ParameterConversion.Unpack<TArg0>(args, 0),
                         ParameterConversion.Unpack<TArg1>(args, 1)),
                     guard,
-                    guardDescription);
+                    guardDescription != null ? guardDescription : guard?.Method.Name);
             }
 
             /// <summary>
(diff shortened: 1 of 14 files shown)


nblumhardt:
Hi Carlos - this looks great, thanks. Sorry about the slow response, still wrapping my head around this as it has been a while since I looked over the full API :-)

Would `guardDescription` be a better name for the `description` arguments in various places?

Also, is it possible to require `description`s internally to always be non-null whenever a `guard` is supplied, and substitute in the default value of the method name at the very outside layer of the API, e.g. in the `IgnoreIf()` method and those like it?

Thanks!

cmendible:
Hi

Parameters could be remamed to guardDescription and actionDescription depending on the use in triggers ot on entry or exit actions. What do you think?

About the second thing: do you mean making descriptions a requiered parameter for methods in the StateRepresentation class?

Thanks!

cmendible:
ok so I renamed the parameters as you can see above!

For the second request I can make the "description parameters" required (not optional) and use the ```Enforce``` class to check for null values as in the following example:
```
public void AddEntryAction(Action<Transition, object[]> action, string description)
            {
                Enforce.ArgumentNotNull(description, "description");
                _entryActions.Add(
                    new EntryActionBehavior(
                        Enforce.ArgumentNotNull(action, "action"),
                        string.IsNullOrEmpty(description) ? action.Method.Name : description));
            }

**Comment by cmendible** [comment]
Ok so in the commits above you can see I renamed the "description" parameters.

As for the second request I can make the "description" parameters required as in the following example:

            public void AddEntryAction(Action<Transition, object[]> action, string description)
            {
                Enforce.ArgumentNotNull(description, "description");
                _entryActions.Add(
                    new EntryActionBehavior(
                        Enforce.ArgumentNotNull(action, "action"),
                        string.IsNullOrEmpty(description) ? action.Method.Name : description));
            }

And that would require a change to the ```StateRepresentationFixture``` to provide not null descriptions.

Also you can check that ```EntryActionBehavior```, ```ExitActionBehavior``` and ```TriggerBehaviour``` classes all replace the "description" with the methods name if the "description" is empty. 

Is that ok with you?

Thanks!

**Comment by nblumhardt** [comment]
Changing the fixture sounds like it would be required, it's not a public API change though - correct?

Nice job on `EntryActionBehavior` and so-on. These are classes I'd expect to require non-null descriptions, which would be computed at the point the behavior object is constructed. Does that match what you're thinking?

I am not sure my last comment implied it, but I'd also aim to replace the default values like `description = ""` with `description = null` since the default description would be calculated at that point (no need to internally pass empty strings around).

Hope this makes sense - thanks!

**Comment by cmendible** [comment]
Absolutely right about the fixtures. 

I think commit 1f4ac98 has everything you expect!

thanks!

File: Stateless/StateConfiguration.cs [hunk 1]
```diff
@@ -372,11 +424,12 @@ public StateConfiguration PermitDynamic<TArg0>(TriggerWithParameters<TArg0> trig
             /// that the trigger will cause a transition to.</param>
             /// <param name="guard">Function that must return true in order for the
             /// trigger to be accepted.</param>
+            /// <param name="guardDescription">Guard description</param>
             /// <returns>The reciever.</returns>
-            public StateConfiguration PermitDynamicIf(TTrigger trigger, Func<TState> destinationStateSelector, Func<bool> guard)
+            public StateConfiguration PermitDynamicIf(TTrigger trigger, Func<TState> destinationStateSelector, Func<bool> guard, string guardDescription = "")
```
nblumhardt: Might have missed a couple of `= null` defaults; this one and the few below use the empty string, which won't result in the method name being used as a default. Cheers!

cmendible:
There we go!

tlk:
I just wanted to add that as a user, you can already use named methods instead of anonymous methods.

Example:

```csharp
machine.Configure(State.A)
    .PermitIf(Trigger.X, State.B, () => true);
```

The corresponding dotgraph will be like this:

```
digraph {
 A -> B [label="X [b__5_0]"];
}
```

`b__5_0` is not very meaningful to someone looking at a diagram. However, if changed to this:

```csharp
bool IsValid() 
{
    return true;
}

    ...
    machine.Configure(State.A)
        .PermitIf(Trigger.X, IsValid);
```

Then the dotgraph contains the method name of the guard, and the resulting diagram is much more readable:

```
digraph {
 A -> B [label="X [IsValid]"];
}
```

The "WhenDiscriminatedByNamedDelegate" test case covers this: https://github.com/nblumhardt/stateless/blob/master/Stateless.Tests/DotGraphFixture.cs#L64

Cheers!

cmendible:
Sure! that was already working for triggers but not for entry and exit actions (the main reason we worked on this changes)

Do you think the code is ready for a merge?

Thanks!

tlk:
Alright, great.

Regarding entry/exit actions I haven't used them and perhaps I am overthinking the issue... but would it be easier to read the graph if the node is altered instead of adding box shaped nodes for each entry and exit action? Perhaps something along the lines of https://en.wikipedia.org/wiki/UML_state_machine#Entry_and_exit_actions

Example dotgraph:
```
digraph {
 A -> B [label="X [guard1]"];
 B [shape=Mrecord,label="{B | entry / Just entered | exit / Just exited}"];
 B -> C [label="Y [guard2]"];
}
```

It might be a little more tedious to generate though, so perhaps an idea for a future improvement?

http://www.graphviz.org/doc/info/shapes.html#record

cmendible:
I agree that the boxes are not the best solution but definitely we should try a better aproach for a future release... If that's ok with @nblumhardt

nblumhardt:
Looking good! Thanks @cmendible and @tlk.

cmendible:
Great! Thanks!

tlk:
I see two approaches to remedy this:
1) add overload methods to be backwards compatible
2) bump the library semver (not sure how that is done with NuGet)

It looks like some people have been struggling with this PR due to the usage of new C# syntax:
- https://github.com/vilcans/stateless/commit/ae14ecd7b65fc1382e6618a40d22c229d2d13ec3
- https://github.com/UnitedTraders/stateless/commit/5cc0fe5b7c8a5696b729d3f76543e2f7887dbb1a

@cmendible @nblumhardt Thoughts?

cmendible:
Maybe it's a silly question but why not just build the project again if you are upgrading the nuget packages?

I understand the issues with the new C# syntax but since [support has ended for .Net 4, 4.5 and 4.5.1](https://blogs.msdn.microsoft.com/dotnet/2015/12/09/support-ending-for-the-net-framework-4-4-5-and-4-5-1/) should we go back? 

@nblumhardt thoughts?

nblumhardt:
Hey folks,

I have to confess a bit of a fast-and-loose approach to SemVer here. I think a binary breaking change is bearable within a major version of this project if source compatibility is maintained, but it's not ideal for libraries that are dependent on Stateless as @borrrden has pointed out. This might be one for more consideration next time around.

The C# version is more of a build-time than runtime issue, so long as you have VS2015 you can target any version of .NET using C#6 so I don't see a lot of value in going back here.

borrrden:
@cmendible to answer your question, say you have a library on Nuget that is dependent on version 2 or greater of a dependency (let's call it dep A). If a third party installs your library they will get both your library and dep A. Now say that dep A makes a binary breaking change in 2.1 and the user who installs your library upgrades dep A. Suddenly you crash at runtime.  You could quickly recompile and release a new version of your library but you need time to test the new version etc.

cmendible:
Sure I get it and alos agree with @nblumhardt. But I don't see the point of upgrading dep A or any dep without first taking the time to test the new version and ensure full compatibilty...

borrrden:
I don't either, but it causes tickets to come in when users accidentally hit the update all button and the stack trace comes through our library last. In all likeliness it would have been a smooth upgrade aside from the missing method.

cmendible:
:+1:

Issue #11, opened 2014-12-01
Firing triggers from OnEntry/OnExit actions.


ianwall:
This is something I do quite frequently - an OnEntry action either directly or indirectly results a call to Fire() that takes the state machine to a new state.

The problem is that because the call to Fire immediately causes a new transition, any code after the call to Fire in the OnEntry action is called _after_ the state machine enters the new state. Here's an example:

```
class Program
{
    enum Trigger
    {
        T1, T2, T3
    }

    enum State
    {
        S1, S2, S3
    }

    static void Main(string[] args)
    {
        var check = true;

        var sm = new StateMachine<State, Trigger>(State.S1);

        sm.Configure(State.S1)
            .Permit(Trigger.T1, State.S2)
            .OnEntry(() => Console.WriteLine("State S1 OnEntry action"));

        sm.Configure(State.S2)
            .Permit(Trigger.T2, State.S3)
            .Permit(Trigger.T3, State.S1)
            .OnEntry(() =>
                {
                    Console.WriteLine("State S2 OnEntry action started");
                    if (check)
                        sm.Fire(Trigger.T2);
                    else
                        sm.Fire(Trigger.T3);
                    Console.WriteLine("State S2 OnEntry action complete");
                });

        sm.Configure(State.S3)
            .Permit(Trigger.T1, State.S1)
            .OnEntry(() => Console.WriteLine("State S3 OnEntry action"));

        sm.Fire(Trigger.T1);
    }
}
```

This results in the output:

State S2 OnEntry action started
State S3 OnEntry action
State S2 OnEntry action complete

I.e. S2 completes after S3.

I have a simple fix for this that add an internal queue to the state machine so that processing of triggers is carried in strict order.

Do you want the fix?

StevenTCramer:
Ian I would love to see this fixed.

HenningNT:
I like to wrap the state machine in an Active Object, that provides an Interface to the world. The outside would not deal with the state machine, but use the public methods to post triggers in a queue. A timer empties the queue at regular intervals, so that all triggers are dealt with in order and atomically.

All trigger handlers also uses the queue, they don't Fire triggers. So I would have something like this for the OnEntry:

            .OnEntry(() =>
                {
                    Console.WriteLine("State S2 OnEntry action started");
                    if (check)
                        _eventQueue.Enqueue(Trigger.T2);
                    else
                        _eventQueue.Enqueue(Trigger.T3);
[...]

PR #47, opened 2016-02-12
Add internal event queue.

ianwall:
Ensure events are fired in the same order they are generated (e.g. from OnEntry/Exit transitions).

This change modifies InternalFire to add the trigger to a queue, and then process it after the current transition is completed (if any).

### DIFF
--- a/Stateless/StateMachine.cs
+++ b/Stateless/StateMachine.cs
@@ -18,6 +18,15 @@ namespace Stateless
         Action<TState, TTrigger> _unhandledTriggerAction = DefaultUnhandledTriggerAction;
         event Action<Transition> _onTransitioned;
 
+        private class QueuedTrigger
+        {
+            public TTrigger Trigger { get; set; }
+            public object[] Args { get; set; }
+        }
+
+        readonly Queue<QueuedTrigger> _eventQueue = new Queue<QueuedTrigger>();
+        private bool _firing;
+
         /// <summary>
         /// Construct a state machine with external state storage.
         /// </summary>
@@ -169,7 +178,34 @@ namespace Stateless
             InternalFire(trigger.Trigger, arg0, arg1, arg2);
         }
         
+        /// <summary>
+        /// Queue events and then fire in order.
+        /// If only one event is queued, this behaves identically to the non-queued version.
+        /// </summary>
+        /// <param name="trigger">  The trigger. </param>
+        /// <param name="args">     A variable-length parameters list containing arguments. </param>
         void InternalFire(TTrigger trigger, params object[] args)
+        {
+            _eventQueue.Enqueue(new QueuedTrigger{Trigger = trigger, Args = args});
+            if (_firing)
+                return;
+
+            try
+            {
+                _firing = true;
+                while (_eventQueue.Count != 0)
+                {
+                    var queuedEvent = _eventQueue.Dequeue();
+                    InternalFireOne(queuedEvent.Trigger, queuedEvent.Args);
+                }
+            }
+            finally
+            {
+                _firing = false;
+            }
+        }
+
+        void InternalFireOne(TTrigger trigger, params object[] args)
         {
             TriggerWithParameters configuration;
             if (_triggerConfiguration.TryGetValue(trigger, out configuration))


Stateless/StateMachine.cs
```diff
@@ -18,6 +18,15 @@ public partial class StateMachine<TState, TTrigger>
         Action<TState, TTrigger> _unhandledTriggerAction = DefaultUnhandledTriggerAction;
         event Action<Transition> _onTransitioned;
 
+        private class QueuedTrigger
```
nblumhardt: Explicit `private` modifier not needed here (just to align with the other code)

Stateless/StateMachine.cs
```diff
@@ -18,6 +18,15 @@ public partial class StateMachine<TState, TTrigger>
...
         event Action<Transition> _onTransitioned;
 
+        private class QueuedTrigger
+        {
+            public TTrigger Trigger { get; set; }
+            public object[] Args { get; set; }
+        }
+
+        readonly Queue<QueuedTrigger> _eventQueue = new Queue<QueuedTrigger>();
+        private bool _firing;
```
nblumhardt: Also no need for `private`.

Stateless/StateMachine.cs
```diff
@@ -169,8 +178,35 @@ public void Fire<TArg0>(TriggerWithParameters<TArg0> trigger, TArg0 arg0)
...
+                _firing = true;
+                while (_eventQueue.Count != 0)
+                {
+                    var queuedEvent = _eventQueue.Dequeue();
+                    InternalFireOne(queuedEvent.Trigger, queuedEvent.Args);
+                }
+            }
+            finally
+            {
+                _firing = false;
```
nblumhardt: Clearing the queue here seems like a worthwhile addition.

nblumhardt:
This looks good, thanks for the PR. I'm not sure how much of a breaking change this might be for existing apps - so I'm slightly uneasy about changing it subtly like this. Have you got a sample scenario that shows this in context?

HenningNT:
I would suggest to think this over a bit more before merging.
In some situations one would have to have LIFO as well as FIFO proccessing, and the proposed solution only supports FIFO. I've wrapped my state machines in active objects, and implemented the queue in the active objects.

Stateless/StateMachine.cs
ianwall: Both true.

Stateless/StateMachine.cs
ianwall: I'm unclear why clearing the queue does anything useful? And exactly where do you mean?

Also, for an example of usage, see my original issue https://github.com/dotnet-state-machine/stateless/issues/11

ianwall:
I'm curious as to when you would want a LIFO queue?

HenningNT:
An example the need for LIFO from my own experimenting With stateless:
I used an state machine to decode bytes received from the Serial port, which assembled message objects. The bytes received were posted to the queue, and the state machine would wait for the start-of-message token, and then post a token-detected event. This event was queued after the other byte-received events. The problem I had was that the other bytes would then be dropped until the token-detected event had been proccessed. I solved this by posting the token-detected event LIFO.

I'm trying using stateless with the Active Object design pattern:
http://www.drdobbs.com/parallel/prefer-using-active-objects-instead-of-n/225700095

roderickgreen:
I just had to throw away a design that used Stateless because of the lack of support for run-to-completion. In my use case, I had to call external code via event handlers, and that code could cause a change of state before my entry handler had completed. I'd like to see this merged as is.

Stateless/StateMachine.cs
Devoney: "Ensure events are fired in the same order " by events do you mean triggers? I'm not sure why there should be a queue on triggers. Isn't it by design of a state machine that it transitions from state to state, and each state evaluates whether a particular trigger makes sense? What problem does the queue solve?

nicolasdev:
This PR is a nice enhancement, I would like a merge too!

HenningNT:
Hi @nicolasdev , what you need is to wrap the state machine in an Active Object (AO). The AO would then expose the public methods needed for the class functionality. Each of the public methods would post an event to its event queue and immediately return. The AO would run a timer and process each event by Fire-ing the event to the Stateless. This ensures both **run-to-completion** and **in-order** event handling.
If you need to trigger new events in the Entry or Exit, post the events to the queue instead of using Fire.

I understand why some want this to be merged. It does solve a problem people are facing. But I think the problem is solved the wrong way. The proposed solution adds complexity and state variables.

nblumhardt:
Merged. Thanks everyone who participated in this thread, sorry it took so long to finalize!

In related news, I've now triaged all PRs and issues, the only ones now remaining open are those we're actively working on. If anyone can help get those few across the line, any and all help appreciated.

Cheers!

Issue #45, opened 2016-02-09
StateConfiguration does not expose configured State and the StateMachine itself


beachwalker:
The current implementation of the **StateConfiguration does not expose the StateMachine that is currently configured. Not exposing the StateMachine makes it difficult and is a limitation for enriching the fluent interface using Extension Methods** without changing the Stateless library itself (you have to add the current StateMachine as parameter which makes some "noise" that shouldn't be required).

This is an example of an Extension Method if you have a Trigger that flows along several states and should be automatically triggered/fired when reaching a state:

```csharp
public static class StateMachineExtensions
{
	public static StateMachine<TState, TTrigger>.StateConfiguration RepeatFireFor<TState, TTrigger>(
		this StateMachine<TState, TTrigger>.StateConfiguration configuration,
		TTrigger trigger,
		IEnumerable<TState> destinations,
		StateMachine<TState, TTrigger> stateMachine)
	{
		return configuration.OnEntry(
			transition =>
				{
					if (!transition.IsReentry && destinations.Contains(transition.Destination))
					{
						stateMachine.Fire(trigger);
					}
				});
	}

	public static StateMachine<TState, TTrigger>.StateConfiguration RepeatFireFor<TState, TTrigger>(
		this StateMachine<TState, TTrigger>.StateConfiguration configuration,
		TTrigger trigger,
		TState destination,
		StateMachine<TState, TTrigger> stateMachine)
	{
		return RepeatFireFor(configuration, trigger, stateMachine, destination);
	}

	public static StateMachine<TState, TTrigger>.StateConfiguration RepeatFireFor<TState, TTrigger>(
		this StateMachine<TState, TTrigger>.StateConfiguration configuration,
		TTrigger trigger,
		StateMachine<TState, TTrigger> stateMachine,
		params TState[] destinations)
	{
		return RepeatFireFor(configuration, trigger, destinations, stateMachine);
	}
}
```

This results in something like this:

```csharp
stateMachine.Configure(State.SomeState)
	.Permit(Trigger.GotoNext, State.FirstNextState)
	.RepeatFireFor(Trigger.GotoNext, State.SecondNextState, stateMachine)
	.OnEntry(DoSomethingOnEnty);
```

Which could be much cleaner and less error prone (client is forced, could insert the wrong state machine):

```csharp
stateMachine.Configure(State.SomeState)
	.Permit(Trigger.GotoNext, State.FirstNextState)
[...]

PR #49, opened 2016-02-24
#45 tried to improve adding of extension method

beachwalker:
Now the currently configured state machine and the configured state are
accessible through properties. This way you can use them within
extension methods without additional temporary variables and method
parameters. See issue #45.


nblumhardt:
Looks good, thanks.

PR #52, opened 2016-03-07
Optimizing InternalFire by reducing calls to the state accessor.

psibernetic:
I found firing a trigger in my system required querying the database 4 times before state could be transitioned and during investigation found non local calls to the State property.

This pull fixes those for that single call.


nblumhardt:
Thank you, looks good!

PR #56, opened 2016-03-18
+ Added -Throw exception containing guard description when no valid t…

### LINKED ISSUE #55: No mention of unmet guard conditions when trigger fire fails in default error handler
It would be nice if the default exception thrown when a trigger is valid for a given state but a guard condition is not met mentioned the guard description text.

Somewhat related to #54

rustybox:
…ransitions due to unmet guard conditions

Closes #55


nblumhardt:
Very nice, thanks Paul!

Issue #59, opened 2016-05-18
Can we have InternalTransition, please?


HenningNT:
I have used the Quantum Leaps state machine framework quite a bit to run state machines on microcontrollers. It supports internal transitions, and it makes life a lot easier. The internal transition does not actually change state, so the state is the same after the event has been processed.

Consider the telephone call example. Let's say we need to adjust the volume. Every state would need an two extra lines: PermitReentry (AdjustVolumeEvent) and OnEntryFrom(AdjustVolumeEvent, HandleVolumeControl).

By using internal transition we only need to add a handler to a superstate (State.OffHook), which is much less verbose.
As a bonus the super states becomes much more useful as well.

nblumhardt:
Hey @HenningNT  - interesting idea! Definitely open to a PR along these lines, so long as existing semantics/backwards compat are maintained. I have a bunch of other commitments that mean I won't be able to tackle this myself, so marking 'up-for-grabs' in case you or someone else watching the project is interested to have a go at it. Thanks!

HenningNT:
I tried to do something with this...
I think I did a pull request, but I am very much a github noob...

Cussa:
Any position about this? I think that it's very useful!

nblumhardt:
@Cussa I'm still open to the idea, but the linked PR is incomplete/not building so I haven't looked any further into it.

HenningNT:
Hi, I have done a bit on this, and I have something that works for me ;)
I haven't gotten round to uploading it yet, I'll try to do that soon.

HenningNT:
Hi!
I have made a pull request, please have a look and see if it works for you :-)

It works for me and my needs ;-)

tlk:
@HenningNT Neat! I have added some minor comments to the PR #73 

Considering the ToDotGraph support, would you expect these transistions to be rendered on a diagram? If so, how?

HenningNT:
I haven't considered dotGraph support yet, but I do know that dotgraph can be used to draw uml charts, and internal transitions are part of uml so I expect that the ToDotGraph method can be modified to draw internal transitions.

So it's a matter of finding someone to do it 😊
I'll take a look at how the diagram is made, but I won't promise anything.

nblumhardt:
I am not sure what the default behaviour will be, but we do probably need to make sure that the rendering in dotGraph is sane. Nothing fancy required though :-)

Otherwise +1 on the PR, thanks for pushing it forward @HenningNT!

[...]

PR #60, opened 2016-05-19
Attempting to do #59, Internal Transitions

HenningNT:
Added support for internal transitions.
Internal transitions does not change the state, and does not execute Exit and Entry actions.

Added use example to Telephone call example.


HenningNT:
I have no idea of how to use GitHub...

HenningNT:
Driving home I realized that this will not work for more than one trigger.
The triggers and corresponding actions need to stored in a Dictonary, or something similar.

PR #63, opened 2016-06-03
Possibility for infinite transitions via decorator pattern.

Devoney:
**Short off-project intro:**
This is my first contribution to an open source project ever. I would like you to consider yourself to be a sort of mentor for when I violate 'common sense' contribution rules.

**What is added:**
The StateMachine is now capable of endless transitions if decorated with the InfiniteTransitionDecorator. For decoration to be possible I've added an interface IStateMachine and an abstract decorator base. I've chosen for the decoration pattern to not interfere with the StateMachine itself, but to put the functionality on top. So the StateMachine is untouched. A user can choose to use the infinite decorator in return for the limitation of not firing more triggers after one another in the same method scope. All Fire methods (thus including those with parameters) are supported.

Construction example:
`
IStateMachine<State, Trigger> stateMachine = new StateMachine<State, Trigger>(State.A);
stateMachine = new InfiniteTransitionsDecorator<State, Trigger>(stateMachine);
`

**Tests:**
The behaviour of the infinite decorator is tested in 2 unit tests up to short.MaxValue transitions. And also the tests in StateMachineFixture are altered to be able to run them using the infinite decorator, and without. (See the inheritance of the TestFixtures).
image

**Dot Graph:**
I've also added the option to leave out ignored triggers in the graph. Which I find convenient. Also unit tested.

I hope you see value in these additions and that these commits will be pulled into this project.
Thank you for creating Stateless.


nblumhardt:
Hi @Devoney - thanks for the PR!

Can you please include an example scenario showing usage of the new code (ideally pretty close to real-world), and a brief note on why the scenario can't be addressed by the codebase today?

(The PR is very much appreciated, just regarding your request for tips: sending the scenario description ahead of doing the work is usually worthwhile, since sometimes although it's a valid scenario, it may not always affect enough users to warrant bringing in the code vs. maintaining a one-off private fork, etc. etc.)

Cheers!

Devoney:
Hi @nblumhardt, thanks for the quick reply. I can understand if this isn't taken into the master branch and a personal fork is more approriate. After some reading up I see indeed that this doesn't add a new possibility. Endless transistions are indeed already possible. It then comes down to the style a statemachine running endlessly is setup.

The following line deepens the callstack, using the InfiniteTransitionsDecorator, it doesn't.

`.OnEntry(() =>
                {
                    stateMachine.Fire(Trigger.Z);
                })`

So an Action could be set on entry, Triggering in that action without deepening the callstack.

```
            stateMachine.Configure(State.A)
                .OnEntry(StateA)
                .Permit(Trigger.X, State.B);
```

So it boils down to something of a preferred way of setting up the StateMachine, with its rendered graph, and indeed not something that wasn't already possible.

Thanks again for the quick reply.

nblumhardt:
Thanks for the follow up Mike. Still interested to learn more about the scenario, but sounds like if the current API can be used successfully for this we would probably stick with it.

Publishing a separate package with the functionality, that depends on and wraps this package, is also a good avenue to explore as it provides an opportunity to test out the API and get feedback?

Devoney:
Yes that is a good idea. In that case, would it be possible to directly put the IStateMachine interface on StateMachine? Or on a derived class in the wrapping package? I will setup a nuget package somewhere this week. A college requested an audit trail on the StateMachine. I am pursuing this in the same decoroation style, logging each public API callThanks for the reply.

nblumhardt:
Hi Mike,

I'm not specifically against an interface, but not sure if it'd be the hook of choice for auditing; have you checked out `OnTransitioned()`?

```csharp
stateMachine.OnTransitioned(t => /* Audit details of t */);
```

Devoney:
Hi @nblumhardt,
Yes I've seen it, thanks. It will be part of the audit trail. What we also want to include in the audit trail is the usage of the API of StateMachine with all its parameters, etc. The audit trail will become somewhat verbose. We use the StateMachine in a multithreaded situation, and it helps us to debug what goes wrong.

nblumhardt:
Interesting, thanks for the information. Could we add more information to `Transition` to provide parameter information? Seems like a worthwhile improvement if it is possible, as the aim of `OnTransitioned()` is to produce such an audit trail :+1:

nblumhardt:
Closing for now as I think we've moved past the attached PR, but please let me know if you're still investigating options here. All the best!

PR #65, opened 2016-06-30
Fix bug where an Ignored trigger shows up as a unknown destination in the dot graph

peter172sp:
Instead, show this a a transition back to the current state.

### DIFF
--- a/Stateless.Tests/DotGraphFixture.cs
+++ b/Stateless.Tests/DotGraphFixture.cs
@@ -211,5 +211,22 @@ namespace Stateless.Tests
 
             Assert.AreEqual(expected, sm.ToDotGraph());
         }
+
+        [Test]
+        public void TransitionWithIgnore()
+        {
+            var expected = "digraph {" + System.Environment.NewLine
+                         + " A -> A [label=\"Y\"];" + System.Environment.NewLine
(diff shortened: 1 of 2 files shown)


nblumhardt:
Thanks for the fix; I think this is better than the current behaviour, but does lead to an ambiguity between "ignored" and "reentry". I will merge as-is, but suggestions for resolving the ambiguity (perhaps labeling as "Y (ignored)" would be welcome.

peter172sp:
Interesting. I didn't even notice the PermitReentry (still learning this library). For my state machine either Ignore or PermitReentry would work. My entry and exit actions on the state in question are just doing some logging, so they don't really affect the current state of the system. 

From a UML state diagram point of view I think a state machine stays in a given state until an event occurs that triggers a state transition. Which means that there is an implied behavior of ignoring other events that don't apply to the current state.

The "Ignore" state configuration seems to be less about the state machine model, and more about giving the user of this library some control over what to do with unhandled triggers. It looks like the default behavior is to throw an exception from the Fire method unless that trigger is explicitly ignored, but this doesn't change the state of the state machine.

Now with this understanding of the library, my advice is to just leave the ignored triggers out of the graph. If I want to show a transition back to myself I should just use PermitReentry. See pull request #66

PR #66, opened 2016-07-01
Do not include ignored triggers in the dot graph

peter172sp:
since both ignored and unhandled triggers do not cause state transitions to occur.


nblumhardt:
Great, think this is also a good representation of what it means to ignore a trigger :+1:

PR #12, opened 2014-12-09
Add portable profile 259 (.NET 4.5) and Xamarin support

anjdreas:
Profile 259 has become more or less the standard PCL profile when targeting .NET 4.5 in the nuget community, and is well supported by the Xamarin framework.

I'm not familiar with signed assemblies, so I'd appreciate a hand on that part.

PR #69, opened 2016-07-15
Add portable profile 259 (.NET 4.5) and Xamarin support (nuget pack .nuspec)

anjdreas:
Second attempt, trying to fix build issues in #12.

### DIFF
--- a/Stateless.sln
+++ b/Stateless.sln
@@ -21,6 +21,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = ".nuget", ".nuget", "{9DBF06
 		.nuget\packages.config = .nuget\packages.config
 	EndProjectSection
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "StatelessNet45Profile259", "Stateless\StatelessNet45Profile259.csproj", "{DC4DD764-F902-4D2F-B175-37EB3A8F006D}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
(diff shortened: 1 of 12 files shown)


nblumhardt:
Looks like CI is still having issues. I've found a few things can fix this, e.g. adding the configuration info to appveyor.yml, but it's possibly the .sln file just needs to be manually edited to specify a different VS version. HTH!

anjdreas:
Reverted solution changes now. And cleaned up things a bit.

anjdreas:
Build succeeds now, hooray :-)

anjdreas:
Not sure about the nuget generator stuff though, also signing the project. Could you help you on that?

anjdreas:
Build seems to be close to good now, but there is a third (duplicate) folder in the .nupkg, which I believe is due to how you call `nuget pack Stateless.csproj` instead of using the `.nuspec` file to pack the nugets.

https://ci.appveyor.com/api/buildjobs/94mc9ltkthq2hyuy/artifacts/Stateless.2.5.76.0.nupkg

nblumhardt:
Great progress!

There also now seems to be an additional fourth part in the NuGet package version (`.0`) that we should drop off.

Just a thought, since the API is unchanged between the versions, can we simplify this by converting to the single, broadly-compatible Profile259 target? (Trying to recall whether we'd drop any platforms that way...)

anjdreas:
It seems other packages created before this PR also has the fourth number, so I think that is unrelated to this PR. Might be an update of the nuget tool in appveyor? 

https://ci.appveyor.com/project/DotnetStateMachine/stateless/build/2.5.68/artifacts

As for unifying to a single project, that is up to you. You will lose .NET4 and SL5 support. If it were up to me, I would support both as I don't see any good reasons to drop support. Both profiles are already handled in this PR.

anjdreas:
What remains to do, is to prevent the third output folder in the .nupkg. There is a duplicate profile 136, likely due to how you are calling nuget pack in the appveyor setup as the folder names are only slightly different. I'm not entirely sure.

The nuspec defines both profiles by folder names, similar to how I have set up one of my other projects:

https://github.com/anjdreas/UnitsNet/blob/master/Build/pack-nuget.bat
https://github.com/anjdreas/UnitsNet/blob/master/Build/UnitsNet.nuspec

anjdreas:
If you call `nuget pack Stateless.nuspec` intsead of `nuget pack Stateless.csproj`, I might have to add back the commit about including source files in the .nuspec.

nblumhardt:
Ah I see - the version number embedded in the package is correct, but the filename includes an extra trailing `0`. Shouldn't be a problem :+1: 

Regarding the build infrastructure, the new `dotnet` tooling should resolve this later in the year, so unless the added folder causes problems at installation-time we can probably ignore it for now. What do you think, time to hit the button?

anjdreas:
It's fine by me, really, but curious got the best of me.
I tried reproducing both build configurations locally and were able to get rid of the third output.

Here is what I propose:
1. I added back "src" in .nuspec
2. You change your nuget pack command from `nuget pack "Stateless\Stateless.csproj" -Properties "Configuration=Release;Platform=AnyCPU" -Symbols` to `nuget pack Stateless\Stateless.nuspec -Properties "Version=<appveyor variable for version>"`

I found this:
http://help.appveyor.com/discussions/questions/1357-nuget-deployment-for-multiple-project-files

Seems like you substitute a nuget build step with a command line build step.

If you want to go with this, change your appveyor setup to make this PR build like you want.
If you want to go with the current setup, and one duplicate output folder in the nuget, then I've created PR #70 for your convenience.

nblumhardt:
Thanks @anjdreas. I think the longer-term plan is to go for a process more like the one we're using in https://github.com/serilog/serilog - though we're waiting for the .NET tooling dust to settle before drawing any conclusions.

PR #70, opened 2016-07-18
Add portable profile 259 (.NET 4.5) and Xamarin support (nuget pack .csproj)

anjdreas:
Duplicate of #69.
Does not have `src` defined in .nuspec, which is redundant when using `nuget pack` on a `.csproj`.


nblumhardt:
Let's give this a shot.

anjdreas:
Tested OK here, thanks!

PR #73, opened 2016-08-15
Added internal transitions

HenningNT:
Added internal transitions.

### DIFF
--- a/Stateless/StateConfiguration.cs
+++ b/Stateless/StateConfiguration.cs
@@ -55,7 +55,7 @@ namespace Stateless
             /// <returns></returns>
             public StateConfiguration InternalTransition(TTrigger trigger, Action<Transition> entryAction)
             {
-                _representation.AddTriggerBehaviour(new InternalTriggerBehaviour(trigger, () => true));
+                _representation.AddTriggerBehaviour(new InternalTriggerBehaviour(trigger));
                 _representation.AddInternalAction(trigger, (t, args) => entryAction(t));
                 return this;
             }
@@ -68,7 +68,7 @@ namespace Stateless
             /// <returns></returns>
             public StateConfiguration InternalTransition<TArg0>(TTrigger trigger, Action<Transition> entryAction)
             {
-                _representation.AddTriggerBehaviour(new InternalTriggerBehaviour(trigger, () => true));
+                _representation.AddTriggerBehaviour(new InternalTriggerBehaviour(trigger));
                 _representation.AddInternalAction(trigger, (t, args) => entryAction(t));
                 return this;
             }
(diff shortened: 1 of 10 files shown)


Stateless/StateConfiguration.cs
```diff
@@ -47,7 +47,44 @@ public StateConfiguration Permit(TTrigger trigger, TState destinationState)
...
                 return InternalPermit(trigger, destinationState, string.Empty);
             }
-
+            /// <summary>
+            /// 
+            /// </summary>
+            /// <param name="trigger"></param>
+            /// <param name="entryAction"></param>
+            /// <returns></returns>
+            public StateConfiguration InternalTransistion(TTrigger trigger, Action<Transition> entryAction)
```
tlk: Maybe rename to InternalTransition without the extra s :)

Stateless/StateMachine.cs
```diff
@@ -208,6 +208,15 @@ void InternalFire(TTrigger trigger, params object[] args)
...
                 newRepresentation.Enter(transition, args);
             }
+            else
+            {
+                var transition = new Transition(source, destination, trigger);
+
+                CurrentRepresentation.InternalAction(transition, args);
+
+//                if (_onTransitioned != null)
+//                    _onTransitioned(transition);
```
tlk: These comments could be removed, right?

Stateless/StateRepresentation.cs
```diff
@@ -207,6 +238,11 @@ public IEnumerable<TTrigger> PermittedTriggers
                     return result.ToArray();
                 }
             }
+            internal void InternalAction(Transition transition, object[] args)
+            {
+                Enforce.ArgumentNotNull(transition, "transtion");
```
tlk: "transtion" -> "transition"

HenningNT:
I tried committing the fixes with Github Desktop app, but it somehow failed. The app tells me the changes are committed, but the link provided by the app leads to a 404 not found :(

Any ideas?

tlk:
Hi @HenningNT I'm not acquinted with the Github Desktop app, and forgive me if this is silly - but perhaps you need to push the locally committed changes to your github repository? http://stackoverflow.com/questions/2745076/what-are-the-differences-between-git-commit-and-git-push

HenningNT:
Yay!
Thanks, that helped. There's a "Sync" button that does the push.

I should probably just use the Git functionality already present in VS2015...

HenningNT:
What happens now?

nblumhardt:
Nice job! Looks like a great feature.

Just a couple of comments:

 * A few tests laying out how the behaviour should work would be beneficial, especially when hierarchical states are involved
 * Since no transition actually takes place, is the name the right one?

HenningNT:
Thanks!
I'll add some tests to make sure it works 😉

The wikipedia page about uml and state machines refer to these event as internal transitions, so I think the name is ok.

nblumhardt:
Looking good. Ready to go?

HenningNT:
Yes!

Issue #67, opened 2016-07-01
Target .NET Core


DamianReeves:
Now that .NET Core is out Stateless makes sense as a project that targets .NET core

shakdoesgithub:
Are there any plans to support dotnet core?

nblumhardt:
A PR based on the project structure, Build.ps1 and appveyor.yml from https://github.com/serilog/serilog would be very welcome if anyone's able to give it a shot :+1:

ArturKarbone:
I just tried to port to .NET Core really quick. Worked out pretty smoothly. The only showstopper is *.resx resource files. Will try to make a PR this week.

ArturKarbone:
just added a [PR ](https://github.com/dotnet-state-machine/stateless/pull/75)for targeting .NET Core

ArturKarbone:
I've just resumed my work on porting to .NET Core and got to the point where I have 2 .NET Core projects for stateless and tests respectivelly with ability to compile for core and other frameworks (got rid of existing "portable" projects). Build.ps1 and appveryor.yml follow Serilog approach with some little adoptation. Working in a separate branch, planning to update the PR later this week.

nblumhardt:
@ArturKarbone great! Thanks for the update :+1:

ArturKarbone:
@nblumhardt just updated pull request. CI pipline works.

peekmessage:
I was just about to start working on getting stateless to build as a .Net Standard library.  From the looks of it, it can target netstandard1.0 which would cover a bunch of platforms including universal and core apps.  Have you considered .Net standard?

nblumhardt:
This is out as 3.x-dev.

PR #75, opened 2016-08-31
Targeting .NET Core

ArturKarbone:
Added .NET Core support by providing Stateless.Core.xporj and project.json files.

Since project.json and existing stateless csprojects live in the same folder (and they don't play nice together) my idea was to provide standalone Stateless.Core.project.json just for .NET Core project, however due to this issue [Dotnet build' doesn't work with <assemblyname>.project.json](https://github.com/dotnet/cli/issues/3546) I was forced to apply  this [workaround](http://stackify.com/using-both-xproj-and-csproj-with-net-core/) folks from stackify applied porting their projects by adding fake/empty Stateless.project.json and StatelessNet45Profile259.project.json files


psibernetic:
Can we not simply move to only project.json? Is this an issue with PCLs?

nblumhardt:
Thanks Arthur! I also would prefer a full-scale move across, but it's great that we're making progress!

In #67 I mentioned that the Serilog repository has a complete project layout, Build.ps1 and appveyor.yml that we've been able to use sucessfully to target .NET Core and the full framework over there - because there's quite a bit of work in getting versioning, tests and so-on working, I would really like to reuse those scripts wholesale, and they work with the out-of-the-box .NET Core project layout so there shouldn't be too much to tweak (`powershell ./Build.ps1` in the root directory does everything).

Are you keen to push on further in that direction? Would be great to get this repository across, I am a bit low on time but happy to help where I can.

ArturKarbone:
> Can we not simply move to only project.json? Is this an issue with PCLs?

The problem is that this affects non .NET Core csproj files (when they are in the same folder as in our case ). So we have to explicitly specify <assembly name>.project.json so build process knows the related project.

ArturKarbone:
> In #67 I mentioned that the Serilog repository has a complete project layout, Build.ps1 and appveyor.yml that we've been able to use sucessfully to target .NET Core

sure will look into this, Nicholas.

nblumhardt:
> The problem is that this affects non .NET Core csproj files

project.json, in combination with .xproj files, can target the full .NET Framework as well as core, so there'd be no need for .csproj files in there any longer. Multi-targeting from project.json is awesome :-) Check out:  https://github.com/serilog/serilog/blob/dev/src/Serilog/project.json#L19

Stateless.Tests/project.json
```diff
@@ -0,0 +1,23 @@
+{
+  "version": "1.0.0-*",
+
+  "dependencies": {
+    "NUnit": "3.4.1",
+    "dotnet-test-nunit": "3.4.0-beta-2",
+    "Stateless": "2.5.0-*"
```
nblumhardt: Should be: `"Stateless": {"target": "project"}`

Stateless/Stateless.Core.xproj
```diff
@@ -0,0 +1,21 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
nblumhardt: Is this file supposed to be here?

Stateless/Stateless.project.xproj
```diff
@@ -0,0 +1,21 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
nblumhardt: Is this file supposed to be here?

Stateless/project.json
```diff
@@ -0,0 +1,27 @@
...
+    "projectUrl": "https://github.com/nblumhardt/stateless",
+    "iconUrl": "https://storage.googleapis.com/google-code-archive/v2/code.google.com/stateless/logo.png",
+    "requireLicenseAcceptance": false
+  }, 
+  "authors": [ "Nicholas Blumhardt and Contributors" ],   
+  "description": "Create state machines and lightweight state machine-based workflows directly in .NET code",
+  "language": "en-US",
+  "title": "Stateless",
+  "frameworks": {
+    "netstandard1.6": {
```
nblumhardt: Looks like we've dropped a lot of platform targets - using project.json it should be possible to keep NET40+ etc.

appveyor.yml
```diff
@@ -9,15 +9,16 @@ assembly_info:
...
+artifacts:
+- path: artifacts/Stateless.*.nupkg
 deploy:
 - provider: NuGet
   api_key:
     secure: U0b7CQk3G8x+xfzoeGz3o26PyoR4G/DuuJ/G63mMrj+mgrDgV0/JA7/4fqqzO2eH
   on:
-    branch: master
\ No newline at end of file
+    branch: master
```
nblumhardt: Using `branch: /^(master|dev)$/` will allow us to publish pre-release packages.

appveyor.yml
```diff
@@ -9,15 +9,16 @@ assembly_info:
   assembly_version: '{version}.0'
```
nblumhardt: Check out the `install` step in: https://github.com/serilog/serilog/blob/dev/appveyor.yml

Stateless.Tests/project.json
```diff
@@ -0,0 +1,23 @@
...
+  },
+
+  "testRunner": "nunit",
+
+  "frameworks": {
```
nblumhardt: Adding e.g. `net4.6` here as well will make sure the tests are run on both CLRs.

Stateless.Tests/Stateless.Tests.csproj
```diff
@@ -10,7 +10,7 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
```
nblumhardt: Looks like this file should be deleted.

OnOffExample/OnOffExample.csproj
```diff
@@ -10,7 +10,7 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
```
nblumhardt: The examples will need to move to project.json in order to use the version of the code that's built within the solution rather than the NUPKG from NuGet, IIRC.

BugTrackerExample/BugTrackerExample.csproj
```diff
@@ -10,7 +10,7 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
```
nblumhardt: See note for the On/Off example below - requires moving to project.json also.

nblumhardt:
I've now created a `dev` branch in this repository and re-targeted the PR to it. This means that once merged, a `-dev-nnnnn` package will be published to NuGet so that we can give it a run :-)

Stateless/Stateless.Core.xproj
ArturKarbone: will throw it away

OnOffExample/OnOffExample.csproj
ArturKarbone: Agreeing with you. Just wanted to have step by step appoach here.

ArturKarbone:
Just update the PR. A short summary:
- Converted samples to .NET Core projects
- Targeting core/net451/net40 frameworks for Stateless
- Targeting core/net451 frameworks for Stateless.Tests (current nunit runner does not support 4.0)
- Cleaning up some mess: like getting rid of obolete csproj files, etc.

BugTrackerExample/Properties/AssemblyInfo.cs
```diff
@@ -2,35 +2,18 @@
...
 
-// General Information about an assembly is controlled through the following 
+// General Information about an assembly is controlled through the following
 // set of attributes. Change these attribute values to modify the information
 // associated with an assembly.
-[assembly: AssemblyTitle("BugTrackerExample")]
-[assembly: AssemblyDescription("")]
 [assembly: AssemblyConfiguration("")]
-[assembly: AssemblyCompany("Hewlett-Packard")]
+[assembly: AssemblyCompany("Microsoft")]
```
nblumhardt: Probably not necessary changes here.

OnOffExample/Properties/AssemblyInfo.cs
```diff
@@ -2,35 +2,18 @@
...
 
-// General Information about an assembly is controlled through the following 
+// General Information about an assembly is controlled through the following
 // set of attributes. Change these attribute values to modify the information
 // associated with an assembly.
-[assembly: AssemblyTitle("OnOffExample")]
-[assembly: AssemblyDescription("")]
 [assembly: AssemblyConfiguration("")]
-[assembly: AssemblyCompany("<Not Specified>")]
+[assembly: AssemblyCompany("Microsoft")]
```
nblumhardt: As above, probably not the right change to make here.

Stateless.Tests/project.json
nblumhardt: Change still needed here.

Stateless/project.json
```diff
@@ -0,0 +1,30 @@
+{
+  "version": "2.5.0-*",
+  "dependencies": {
+    "NETStandard.Library": "1.6.0"
+  },
+  "packOptions": {
```
nblumhardt: Missing code signing (if we drop signing we'll break existing consumers).

Stateless/project.json
nblumhardt: Can we go back as far as 1.0 or 1.3?

TelephoneCallExample/Properties/AssemblyInfo.cs
```diff
@@ -2,35 +2,18 @@
...
 
-// General Information about an assembly is controlled through the following 
+// General Information about an assembly is controlled through the following
 // set of attributes. Change these attribute values to modify the information
 // associated with an assembly.
-[assembly: AssemblyTitle("TelephoneCallExample")]
-[assembly: AssemblyDescription("")]
 [assembly: AssemblyConfiguration("")]
-[assembly: AssemblyCompany("<Not Specified>")]
+[assembly: AssemblyCompany("Microsoft")]
```
nblumhardt: Probably not the right change here.

Stateless/Properties/AssemblyInfo.cs
```diff
@@ -28,6 +28,6 @@
 [assembly: AssemblyVersion("2.4.0.1")]
 [assembly: AssemblyFileVersion("2.4.0.1")]
 
-[assembly: InternalsVisibleTo("Stateless.Tests, PublicKey=002400000480000094000000060200000024000052534131000400000100010077d5f127d9877b3b813017ddefb168b265f8041ec908984bf36071156ff911a7a27b5243decac72586ecf45edfb93c50fdf063639e39edff2238a2b999caf0b2c9767a1517cb1df747b58853503e03e1fb12f1336c35aa97ad8a8bc2984b7cae9f49cad2eedebb8cf59b3bd5f24fd9f3a53bd1a2d8499f4fe0d938873b0101ca")]
```
nblumhardt: This will need reverting when signing is re-enabled.

appveyor.yml
nblumhardt: Change still needed here (without it we'll use varying .NET Core versions depending on build machine config, IIRC).

TelephoneCallExample/Properties/AssemblyInfo.cs
ArturKarbone: Yeah i belive all that AssemblyCompany("Microsoft") appeared after creating a new xprojects.  I saw that diff before pushing,  but to me it was not a big  issue having "<Not Specified>" versus "Microsoft" for sample project. Will revert that

ArturKarbone: > Can we go back as far as 1.0 or 1.3?
yep

Stateless/Properties/AssemblyInfo.cs
nblumhardt: `[InternalsVisibleTo]` has to have a key specified if it appears in a signed library. With Stateless.dll signed, the tests can't get internal access unless they're also signed.

nblumhardt:
Yikes, the new GitHub review system is a bit confusing to navigate... I hope I've covered everything but let me know if there's something I missed.

ArturKarbone:
@nblumhardt Nicholas: projects are being refrenced using "target" option. In addition to that i've just added/returned assemblies signing. Looks like we went through all the major concernes (still not sure about appveyor config though). Could you make yet anoter review please? Thanks

Stateless/Stateless.project.xproj
nblumhardt: @ArturKarbone I think this is a leftover file from the conversion, is that the case?

BugTrackerExample/project.json
```diff
@@ -0,0 +1,23 @@
...
+  },
+
+  "dependencies": {
+    "Microsoft.NETCore.App": {
+      "type": "platform",
+      "version": "1.0.0"
+    },
+    "Stateless": {
+      "target": "project",
+      "version": "2.5.0-*"
```
nblumhardt: `"version"` doesn't need to be specified in a `"target": "project"` dependency (I don't think it hurts, but it'll get out of sync as the main assembly's version evolves).

Stateless/project.json
```diff
@@ -0,0 +1,31 @@
...
+    "licenseUrl": "http://www.apache.org/licenses/LICENSE-2.0",
+    "projectUrl": "https://github.com/nblumhardt/stateless",
+    "iconUrl": "https://storage.googleapis.com/google-code-archive/v2/code.google.com/stateless/logo.png",
```
nblumhardt: The URL here makes me realize it's time we did a new logo ;-) LOL.

Stateless/project.json
```diff
@@ -0,0 +1,31 @@
...
+    "requireLicenseAcceptance": false
+  },
+  "authors": [ "Nicholas Blumhardt and Contributors" ],
+  "description": "Create state machines and lightweight state machine-based workflows directly in .NET code",
+  "language": "en-US",
+  "title": "Stateless",
+  "buildOptions": { "keyFile": "Stateless.snk" },
+  "frameworks": {
+    "netstandard1.0": {
+      "imports": "dnxcore50",
```
nblumhardt: If I'm not mistaken, this will make all of the dnxcore50 APIs available, which effectively promotes the 1.0 standard up to ~1.5, so chances are it won't run on the very early platforms.

nblumhardt:
Awesome! @ArturKarbone I am happy to merge this when you give the all-clear.

There are a few nitpicks that it would be good to resolve, but I'm also fine with addressing these on `dev` if it's easier for everyone to start collaborating there.

The targeting/netstandard1.0 item is the main one we'll need to give more thought to; it may be that we do whatever we can to get netstandard1.0 supported without any imports, and then version the library up to 3.0 to indicate we may not support all of the earlier profile targets.

Stateless/project.json
ArturKarbone: Exactly, cause nowadays there is no log in nuget package (just nuget default one).

ArturKarbone:
@nblumhardt just resolved main concerns:

- Got rid of dnxcore50 dependency
- Got rid of version when targeting projects (you were right it works with "version" but generates warnigs during compilation when versions are out of sync). 

Regarding tests: dotnet-test-nunit is limiting us to .NETCoreApp 1.0 and 4.5.1. Should not be a major showstopper still. May contribute there as well. 

Looks like we are good to merge.

Tnanks

nblumhardt:
:tada: - thanks for making this happen! I'll keep an eye on the dev build and make sure it spits a package out.

nblumhardt:
Bumped the major version to 3.0, as this will make it less surprising when tooling needs to change. https://www.nuget.org/packages/Stateless/3.0.0-dev-00092

ArturKarbone:
Amazing, Nicholas. I've just created yet another pull request which is going to simplify versions syncing: https://github.com/dotnet-state-machine/stateless/pull/84

PR #84, opened 2016-10-03
Getting rid of assembly patching to simplify versions syncing (Since project.json is used for versioning anyway)

ArturKarbone:


ArturKarbone:
@nblumhardt, this is a small tweak to simplify versioning syncing. The issue now is that the nuget prerelese version is marked as 3.0.0.* however AssemblyVersion was patched as 2.5.*.  I suggest to use project.json for versioning. Thanks

nblumhardt:
Hi Artur, thanks for the PR!

Assembly versioning and project.json versioning are mutually exclusive, so I don't think this will work. Although the NuGet package will end up with the correct version, the assembly itself will end up stamped with 2.4.x or whatever happens to be in the AssemblyInfo.cs file.

Because of the prominence of the assembly version in builds that use strong naming I think we should keep the current setup. What do you think?

Cheers,
Nick

ArturKarbone:
Nick,

Yeah i totally forgot about AssemblyVersionAttribute.  It turned out that .NET Core projects created from scratch don't use AssemblyVersionAttribute at all. Btw. when no AssemblyVersion is specified then project.json's version is uses for assembly versioning.

I see the following options here:

1) Get rid of AssemblyVersionAttribute/Get rid of assemby patching in appveyor.yaml
2) Keep AssemblyVersionAttribute/Get rid of assemby patching in appveyor.yaml/ Introduce [SharedAssembly pattern ](http://stackoverflow.com/questions/6771694/shared-assemblyinfo-for-uniform-versioning-across-the-solution)
3) Get rid of AssemblyVersionAttribute/Stick to assemby patching in appveyor.yaml

To be honest i've never used signed assemblies in my practice. Is it an issue to have different assembly versions signed with the same key?

Btw. current prerelease package use appveyour patching:

nblumhardt:
Thanks for the follow-up.

I don't think we can do 1) - it'd work if we used four-part version numbers, but it doesn't look like SemVer versions are supported: https://github.com/dotnet/cli/issues/3459#issuecomment-225129289

Since we're only shipping a single assembly from the solution, not sure that 2), a _SharedAssemblyInfo.cs_ approach, will win us much.

One other option that might be worth considering - just us a single, static assembly version number. Serilog does this - the version just stays at `2.0.0.0` for all 2.x series builds (https://github.com/serilog/serilog/blob/dev/src/Serilog/Properties/AssemblyInfo.cs).

Signed assemblies cause the CLR to do precise version matching - binding redirects are required even for a very minor version change, otherwise the loader blows up. This is why the un-changing version number strategy works well - it minimises the amount of mucking about with binding redirects :-)

ArturKarbone:
Thanks for the explanation, Nick. let's just leave it as is and leverage appveyor assembly patching then.

nblumhardt:
👍

PR #89, opened 2016-10-29
Async actions support

yevhen:
I like this library so much but can't use it with Orleans due to absence of support for async actions. In Orleans, actors are thread-safe turn-based and everything is based on promises (Tasks) and async/await. So having such support is required to be able to successfully build and execute FSM's inside actor.

Besides `FireAsync` I've added async support to all places where action could be specified:
- `OnUnhandledTrigger`
- `OnTransitioned`
- `OnEntry`
- `OnExit`
- `InternalTransition`

I've used Async suffix to not break binary compatibiilty with previous versions, since consumers may have callbacks specified as method groups and compiler won't be able to resolve ambiguity in such cases.

Also, I added checks for cases when consumer invoke synchronous `Fire` but current transition path involves execution of async actions. This will throw `IOP`. Includes, `OnTransitioned` and `OnUnhandledTrigger` callbacks as well.  It's ok to have sync actions called when using `FireAsync` - in this case completed task will be returned.

NOTE: Async version is not available for NET40 due to absence of async/await support.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ Stateless.userprefs
 .vs/
 
 project.lock.json
+artifacts/
+TestResult.xml
--- a/Stateless.Tests/AsyncActionsFixture.cs
+++ b/Stateless.Tests/AsyncActionsFixture.cs
@@ -1,0 +1,179 @@
+﻿#if !NET40
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+using NUnit.Framework;
+
+namespace Stateless.Tests
+{
+    [TestFixture]
+    public class AsyncActionsFixture
+    {
+        [Test]
+        public async Task CanFireAsyncEntryAction()
+        {
+            var sm = new StateMachine<State, Trigger>(State.A);
+
+            sm.Configure(State.A)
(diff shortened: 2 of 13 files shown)


nblumhardt:
@yevhen wow - very elegantly done. Thanks for the great PR!

I have looked briefly over the code and can't spot any issues - there are quite a few changed lines there so another set of eyes on it would be appreciated if anyone has some time.

Otherwise - LGTM.

Maximusya:
Anticipating the issues/questions arising after someone tries to use a library with words *async* that does not do some magic for the developer: it should be explicitly stated somewhere that **Stateless is still not thread-safe with this PR** (in the README.md in bold or in every async method's summary): in @yevhen case it is the Orleans that makes everything safe.

yevhen:
@Maximusya sure, that should be mentioned in docs along with unavailablity for NET40.

nblumhardt:
@Maximusya great suggestion, thanks. (It is interesting the way that _async_ and _concurrent_ are brought into contrast like this in modern .NET APIs.) I'll make the README update.

Merging now; there should be a new -dev-nnnnn build published directly out of CI.

yevhen:
👍

PR #91, opened 2016-10-31
Initial state activation

yevhen:
One more commit to make Stateless fully compatible with stateful single-threaded actors. 

In Orleans (and other actor frameworks, like Akka.Net), actors are activated by special event, like OnActivateAsync. At that moment previously persisted (in cold storage) state need to be reactivated, since actors are *active* entities. Current approach with specifying initial state via ctor doesn't cut for multiple reasons:

1. The most important - it doesn't enter state (and its superstates) and so OnEntry actions are not called
2. Even if it does - we don't have async ctors in C#

I did few changes to make that possible:

1. Added ability to activate state via special method - `Activate`/`ActivateAsync`. For that I've introduced special case transition `Activation` which doesn't trigger `OnTransitioned` but does trigger `OnEntry` for state and all of its superstates. Also, I've added check to ignore actions specifed via `OnEntry(Action<Transition>)` since its a special transition which doesn't make sense to intercept. Basically, only callbacks specified in parameterless `OnEntry` will run, which is completely inline with semantics of activation.
2. Added checks that state is actually intialized (either via ctor or `Activate`) before any kind of state access happens

In the end:

1. It's fully compatible with previous version. I've only made default ctor public.
2. There now 2 distinct but clear semantics - ctor is for initalizing machine without executing `OnEntry` callbacks defined for initial state (*passive mode*) and `Activate` will execute `OnEntry` callbacks for the given initial state (*active mode*)

Thanks!

### DIFF
--- a/Stateless.Tests/ActiveStatesFixture.cs
+++ b/Stateless.Tests/ActiveStatesFixture.cs
@@ -1,0 +1,200 @@
+﻿using System;
+using System.Collections.Generic;
+
+using NUnit.Framework;
+
+namespace Stateless.Tests
+{
+    [TestFixture]
+    public class ActiveStatesFixture
+    {
+        [Test]
+        public void WhenActivate()
+        {
+            var sm = new StateMachine<State, Trigger>(State.A);
+
+            var expectedOrdering = new List<string> { "ActivatedC", "ActivatedA" };
+            var actualOrdering = new List<string>();
+
+            sm.Configure(State.A)
+              .SubstateOf(State.C)
+              .OnActivate(() => actualOrdering.Add("ActivatedA"));
+
+            sm.Configure(State.C)
+              .OnActivate(() => actualOrdering.Add("ActivatedC"));
+
+            // should not be called for activation
+            sm.OnTransitioned(t => actualOrdering.Add("OnTransitioned"));
+
+            sm.Activate();
+
+            Assert.AreEqual(expectedOrdering.Count, actualOrdering.Count);
+            for (int i = 0; i < expectedOrdering.Count; i++)
+                Assert.AreEqual(expectedOrdering[i], actualOrdering[i]);
+        }
+
+        [Test]
+        public void WhenActivateIsIdempotent()
+        {
+            var sm = new StateMachine<State, Trigger>(State.A);
+
+            var actualOrdering = new List<string>();
+
+            sm.Configure(State.A)
+              .SubstateOf(State.C)
+              .OnActivate(() => actualOrdering.Add("ActivatedA"));
+
+            sm.Configure(State.C)
+              .OnActivate(() => actualOrdering.Add("ActivatedC"));
+
+            sm.Activate();
+
+            actualOrdering.Clear();
+            sm.Activate();
+
+            Assert.AreEqual(0, actualOrdering.Count);
+        }
+
+        [Test]
+        public void WhenDeactivate()
+        {
+            var sm = new StateMachine<State, Trigger>(State.A);
+
+            var expectedOrdering = new List<string> { "DeactivatedA", "DeactivatedC" };
+            var actualOrdering = new List<string>();
+
+            sm.Configure(State.A)
+              .SubstateOf(State.C)
+              .OnDeactivate(() => actualOrdering.Add("DeactivatedA"));
+
+            sm.Configure(State.C)
+              .OnDeactivate(() => actualOrdering.Add("DeactivatedC"));
+
+            // should not be called for activation
+            sm.OnTransitioned(t => actualOrdering.Add("OnTransitioned"));
+
+            sm.Activate();
+            sm.Deactivate();
+
+            Assert.AreEqual(expectedOrdering.Count, actualOrdering.Count);
+            for (int i = 0; i < expectedOrdering.Count; i++)
+                Assert.AreEqual(expectedOrdering[i], actualOrdering[i]);
+        }
+
+        [Test]
+        public void WhenDeactivateIsIdempotent()
+        {
+            var sm = new StateMachine<State, Trigger>(State.A);
+
+            var actualOrdering = new List<string>();
+
+            sm.Configure(State.A)
+              .SubstateOf(State.C)
+              .OnDeactivate(() => actualOrdering.Add("DeactivatedA"));
+
+            sm.Configure(State.C)
+              .OnDeactivate(() => actualOrdering.Add("DeactivatedC"));
+
+            sm.Activate();
+            sm.Deactivate();
+
+            actualOrdering.Clear();
+            sm.Activate();
+
+            Assert.AreEqual(0, actualOrdering.Count);
+        }
+
+        [Test]
+        public void WhenTransitioning()
+        {
+            var sm = new StateMachine<State, Trigger>(State.A);
+
+            var expectedOrdering = new List<string>
+            {
+                "ActivatedA",
+
+                "DeactivatedA",
+                "ExitedA",
+                "OnTransitioned",
+                "EnteredB",
+                "ActivatedB",
+
+                "DeactivatedB",
+                "ExitedB",
+                "OnTransitioned",
+                "EnteredA",
+                "ActivatedA"
+            };
+
+            var actualOrdering = new List<string>();
+
+            sm.Configure(State.A)
+              .OnActivate(() => actualOrdering.Add("ActivatedA"))
+              .OnDeactivate(() => actualOrdering.Add("DeactivatedA"))
+              .OnEntry(() => actualOrdering.Add("EnteredA"))
+              .OnExit(() => actualOrdering.Add("ExitedA"))
+              .Permit(Trigger.X, State.B);
+
+            sm.Configure(State.B)
+              .OnActivate(() => actualOrdering.Add("ActivatedB"))
+              .OnDeactivate(() => actualOrdering.Add("DeactivatedB"))
+              .OnEntry(() => actualOrdering.Add("EnteredB"))
+              .OnExit(() => actualOrdering.Add("ExitedB"))
+              .Permit(Trigger.Y, State.A);
+
+            sm.OnTransitioned(t => actualOrdering.Add("OnTransitioned"));
+
+            sm.Activate();
+            sm.Fire(Trigger.X);
+            sm.Fire(Trigger.Y);
+
+            Assert.AreEqual(expectedOrdering.Count, actualOrdering.Count);
+            for (int i = 0; i < expectedOrdering.Count; i++)
+                Assert.AreEqual(expectedOrdering[i], actualOrdering[i]);
+        }
+
+        [Test]
+        public void WhenTransitioningWithinSameSuperstate()
+        {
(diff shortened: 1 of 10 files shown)


nblumhardt:
Thanks @yevhen. To make sure I understand you correctly - is this only applicable once during a `StateMachine`'s lifecycle?

If so, would this be functionally similar to declaring an additional `Uninitialized` state per statemachine, and then defining `OnEntry()` actions on the transition between `Uninitialized` and the normal 'first' state?

yevhen:
Yes, once. Initial state. Either passive initialization - OnEntry actions will not run, or active - they will.

> If so, would this be functionally similar to declaring an additional Uninitialized state per statemachine, and then defining OnEntry() actions on the transition between Uninitialized and the normal 'first' state?

Ye, probably. But then ctor still doesn't cut since we need to run OnEntry async actions.

yevhen:
>If so, would this be functionally similar to declaring an additional Uninitialized state per statemachine, and then defining OnEntry() actions on the transition between Uninitialized and the normal 'first' state?

The problem with that approach, is that I can move between states and same OnEntry actions need to be executed. There won't be any Unititalized transition then.

nblumhardt:
Thanks for the follow-up. The scheme I had in mind is:

```csharp
enum State { Uninitialized, Initial, ... }
enum Trigger { Activate, ... }
```

In actor ctor:

```csharp
sm = new StateMachine(State.Uninitialized);

sm.Configure(State.Uninitialized)
    .Permit(Trigger.Activate, State.Initial);

sm.Configure(State.Initial)
    .OnEntryAsync(...);
```

Then in the actor's activating event, `sm.FireAsync(Trigger.Activate)`?

I guess the meta-question is, if we create a "virtual" state that the state machine is implicitly in before the proposed "activated" event, how does this differ from representing this as a first-class state?

Since the state machine must always be in _some_ state, even if it's not one the user declares, and Stateless provides a range of facilities for triggering and reacting to transitions, it seems on the surface that the mechanics for modelling the scenario should already be present.

Sorry if I'm on the wrong track here - appreciate your input. Thanks!

yevhen:
> Sorry if I'm on the wrong track here 

You are 😃 

The thing is there is no any actual transition happening. Basically, it's just the way to resume machine in it's last known position (state) - the one it was before the actor was deactivated. Deactivation could happen due to multiple reasons: node crash, GC, etc. When it is activated next time (with client request or by reminder message) it needs to restore the state it was in before it's been GC'ed. With every transition actors store their current state (incl data) in some durable medium. 

I'll give you an example from real-world project (wanrning: it's lengthy):

```cs
public class Task : Actor
{
    readonly ITaskStorage storage;
    readonly StateMachine<State, Trigger> sm = new StateMachine<State, Trigger>();

    enum State
    {
        Initial,
        Cancellable,
        Active,
        Started,
        Running,
        Failed,
        Delayed,
        Completed,
    }

    enum Trigger
    {
        Start,
        Retry,
        Cancel,
        Completed,
        Prepared,
        Executed,
        Delayed,
        Failed,
        Notify,
        Due
    }  

    public Task(string id, IActorRuntime runtime, ITaskStorage storage)
        : base(id, runtime)
    {
        this.storage = storage;
        Configure();
    }

    void Configure()
    {
        sm.Configure(State.Cancellable)
            .Permit(Trigger.Cancel, State.Completed);

        sm.Configure(State.Active)
            .SubstateOf(State.Cancellable)
            .OnEntryAsync(()=> Reminders.Register("keepalive", TimeSpan.FromMinutes(1), TimeSpan.FromMinutes(1)))
            .OnExitAsync(()=> Reminders.Unregister("keepalive"))
            .Permit(Trigger.Completed, State.Completed);

        sm.Configure(State.Initial)
            .Permit(Trigger.Start, State.Started);

        sm.Configure(State.Started)
            .SubstateOf(State.Active)
            .OnEntry(()=> Timers.Register("prepare", TimeSpan.FromSeconds(1), Prepare))
            .Permit(Trigger.Prepared, State.Running);

        sm.Configure(State.Running)
            .SubstateOf(State.Active)
            .OnEntry(()=> Timers.Register("execute", TimeSpan.FromMilliseconds(1), TimeSpan.FromMilliseconds(1), Execute))
            .OnExit(()=> Timers.Unregister("execute"))
            .Permit(Trigger.Failed, State.Failed)
            .Permit(Trigger.Delayed, State.Delayed)
            .InternalTransition(Trigger.Executed, () => {});

        sm.Configure(State.Delayed)
            .SubstateOf(State.Active)
            .OnEntry(()=> Timers.Register("checkdue", TimeSpan.FromMilliseconds(1), TimeSpan.FromMinutes(1), CheckDue))
            .OnExit(() => Timers.Unregister("checkdue"))
            .Permit(Trigger.Due, State.Running);

        sm.Configure(State.Failed)
            .SubstateOf(State.Active)
            .OnEntry(()=> Timers.Register("err", TimeSpan.FromSeconds(1), TimeSpan.FromSeconds(10), NotifyError))
            .OnExit(()=> Timers.Unregister("err"))
            .Permit(Trigger.Retry, State.Running)
            .InternalTransition(Trigger.Notify, () => { });
    }

    public override async Task OnActivate()
    {
        await Load();
        await Reactivate();
    }

    async Task Load()
    {
        Data = await storage.Load() ?? new TaskData();
    }

    async Task Reactivate()
    {
        var state = Data.State != null
            ? (State) Enum.Parse(typeof(State), Data.State)
            : State.Initial;

        await sm.ActivateAsync(state);
    }
```

I also have code which saves Task data (and so current fsm state) after every successfull transition but it's not very interesting.

The question is: if my actor is in `Running` state and it's the last state I've previously saved to disk, then how should set it as initial upon activation, so that its OnEntry callback is executed and OnEntry callback of its parent Active superstate is executed as well?

yevhen:
Perhaps, the missing thing here is some additional special state events, like OnEntry. Something like OnActivate/OnDeactivate. At that point I'm confused how they are different from OnEntry/OnExit. 

The troubling element in current Stateless design is overload of OnEntry which takes Transition as an argument.

nblumhardt:
Ah, I see. By:

>  is this only applicable once during a StateMachine's lifecycle?

I really meant to imply the actor's lifecycle. I should have been clearer, sorry.

`OnActivate`/`OnDeactivate` seem (at first glance) closer to the required semantics than `OnEntry()`, which is based around a transition as you've noted. At present, `OnEntry()` really means "the state machine transitioned into the state", _not_ "the state machine was re-loaded in this state".

Mixing the two would make it hard to incorporate `OnEntry()` handlers with side-effects into a state machine that also depended on `OnEntry()` for post-persistence initialization.

`OnActivated()` plus the `Activate()`/`ActivateAsync()` methods and their de-activation counterparts seem like a solid design - any thoughts on how to incorporate this without making the difference between `OnActivate()` and `OnEntry()` confusing for the user?

yevhen:
> OnActivate/OnDeactivate seem (at first glance) closer to the required semantics than OnEntry(), which is based around a transition as you've noted. At present, OnEntry() really means "the state machine transitioned into the state", not "the state machine was re-loaded in this state".

I agree that `OnActivate/OnDeactivate` pair is more elegant and up to the point design.

> how to incorporate this without making the difference between OnActivate() and OnEntry() confusing for the user?

After some thought I don't think it would be confusing at all. UML differentiates *active* states explicitly. That's exactly the states (Started, Running) I have in exampe above - they do some background work, usually by the timer and they need to be activated so they can setup the timer. So there is nothing to reinvent - we can just state in documentation official terminology and semantics.

I'll take this road and will update PR.

yevhen:
I like this design much more than previous one. Now there is a truly clear concept of *active* state. Thanks for asking right questions!

nblumhardt:
Looks great. I will add some notes on Activate/Deactivate vs. Entry/Exit to the README when I get a chance, unless you beat me to it :-)

beatsm:
@yevhen - Hi I am also looking to use this with Akka.net. Do you have a link to a basic example?

Thanks

yevhen:
@beatsm [this](https://github.com/dotnet-state-machine/stateless/pull/91#issuecomment-257456293) is what I'm using with Orleans actors. For Akka.Net - that will be different. Have you tried using http://getakka.net/docs/Scheduler ?

beatsm:
@yevhen Thanks for the response, I must have got the wrong end of the stick and thought that you had used the changes to hook into the Akka.net persistence life cycle.

I will investigate further.

Thanks

joonhwan:
@beatsm any update? I also considering the integration stateless and akka.net. 
I only have simplest/minimal example code and have shared just a moment ago in the github.
see [here](https://github.com/joonhwan/Akka.Net-Stateless-Demo)

On persistent part, I have yet no idea how to handle it properly using the new `Activate/Deactivate' concept, though.

yevhen:
Load previously saved state [here](https://github.com/joonhwan/Akka.Net-Stateless-Demo/blob/master/Akka-Stateless-Demo/Msm/StatelessHsmActor.cs#L21) and call Activate after you've configured SM.

yevhen:
Your SM wrapper looks more complex than it should be. You just need to map message types to auto-assigned ints using a dictionary and then map them to TriggerWithParameters<TMessage>(TTrigger)

joonhwan:
@yevhen Oh, there i can see a light.  Could I get some more detail code snippet or something relevant to 'auto-assigned ints' mapping? I need a way to define trigger(or event) without defining Enum in a more simple way.

Maximusya:
We have tried to make use of this new **Activate** semantics and... rolled back to plain old Entry/Exit/Transitioned. Either we have not understood the concept or the concept is not finished yet?

Our thoughts:
sm.Activate()/Deactivate() is only called publicly, thus private `this.active` flag is checked/set only twice per state machine lifetime. Transitions via permitted triggers do not touch this flag and if a machine transitions into a state other than the initial one it was activated in - `sm.Deactivate()` would see a state with `this.active == false` and thus would skip any callbacks.

Where is the active state concept then?
Do we have to manually call `sm.Activate()/Deactivate()` in OnTransitioned/OnEntry/OnExit callbacks to preserve a valid value of `this.active` flag?
Or the implementations of `StateRepresentation.Exit/Enter` should call `this.Deactivate()/this.Activate()` instead of `this.ExecuteDeactivationActions()/this.ExecuteActivationActions()` to preserve the active flag in a valid state?

Unit tests added with this PR do not demonstrate the above mentioned edge cases behavior.
Neither they demonstrate the reentry behavior: OnActivate/OnDeactivate for reentry non-nested state transitions. We accepted the concept of **reentrancy**, but we fail to accept the **reactivation** that happens for reentry.

We expect a state to become active once state machine transitions into it from another state or is activated in it.
We expect a state to become inactive once state machine transitions from it to another state or is deactivated while in it.
We do not expect state's OnActivate/OnDeactivate callbacks to be called for reentry transitions, because the state never ceases to be active.

HenningNT:
I believe that most users won't need to use Active/Deactivate, I think the concept was included to conform to concepts used in Akka.NET and/ or Orleans.

If I understand it correctly, activation would do whatever was necessary to get the active object running. An active object might have a timer that empties a message queue. Deactivating the active object could free any resources and stop the timer.

PR #93, opened 2016-11-01
mark stateless as CLS Compliant

SurajGupta:


SurajGupta:
will submit another pull request with instructions on how to build.

PR #96, opened 2016-11-02
Modernize the project layout

nblumhardt:
This is a bit of a monster changeset and sure to cause some downstream merge issues because of the large whitespace change, but I think it's time this work was done. I've wrapped it all up in one giant PR so that hopefully we only need to go through this once :-)

 * Removed remaining old-style NuGet files
 * Removed unused help file builder
 * Re-enabled XML documentation
 * Fixed assembly version patching
 * `/src`, `/test` etc
 * Feature- rather than target-based `#if` defines
 * **CRLF normalization** (*nix-friendly development)


nblumhardt:
@dotnet-state-machine/reviewers just a heads-up, once this is merged to avoid conflict headaches you'll need to update your working copy before making new changes.

PR #98, opened 2016-11-11
Update Bug.cs

shanselman:


nblumhardt:
:-) thanks, Scott!

psibernetic:
Haha. Hanselman, swoops in and erases gender assumptions in the work place. :P Thanks.

nblumhardt:
When I wrote this example in 2008 _I_ was (just about to be) the new employee :-) ...not so much an assumption at that point. Worthwhile change, though. We need a Roslyn-powered plug-in to spot these things at build time...

Issue #80, opened 2016-09-20
SubstateOf and dotGraph


sle118:
First, I want to say that this little state machine project has simplified quite significantly an application I am building for my own fun. I can see the benefits of this approach for future enterprise level projects, although the state machine will likely be an off the shelf solution that exist currently in our landscape.

I'm wondering if it would be possible to have the dot graph represent content inherited from SubstateOf , as it would help picturing more complex state machines. 

I got around the limitation by duplicating all of the "permit" from the parent States, but it defeats the purpose of inheritance and makes the code more complicated to maintain.

Thank you!

HenningNT:
dotGraph doesn't really like substates. It is possible to use HTML inside the states, but I think it would require quite an effort to get right.
It might be easier to export to a different format, one that QM can parse.
http://www.state-machine.com/qm/

One could also make a import function to Stateless to read models from QM at run time, I guess.

sle118:
Or perhaps sub states could be represented as clusters using sdp
http://www.graphviz.org/content/point-edge-cluster

An example here
https://stamm-wilbrandt.de/Viz.js/form10/?1425770108930#

tlk:
@sle118 It would be interesting to see if this would work!

I take it that you meant "fdp" as in http://graphviz.org/content/fdpclust, right?

By the way, this sample link seems to work better for me:
http://stamm-wilbrandt.de/Viz.js/form10/?1425770108930#graph%20G%20%7B%0A%20%20layout%3Dfdp%0A%0A%20%20e%0A%20%20subgraph%20clusterA%20%7B%0A%20%20%20%20a%20--%20b%3B%0A%20%20%20%20subgraph%20clusterC%20%7B%0A%20%20%20%20%20%20C%20--%20D%3B%0A%20%20%20%20%7D%0A%20%20%7D%0A%20%20subgraph%20clusterB%20%7B%0A%20%20%20%20d%20--%20f%0A%20%20%7D%0A%20%20d%20--%20D%0A%20%20e%20--%20clusterB%0A%20%20clusterC%20--%20clusterB%0A%7D%0A

sle118:
Ok. Forget about fdp. After a little more digging, I found what I believe is a better alternative. 

First add the following attribute
`compound=true;`
Next use sub graph clusters to represent nodes that have sub states (the stateless class makes this available) 
`subgraph cluster1 {
e -> g;
e -> f;
}
`
[...]

PR #100, opened 2016-11-11
Use HTML representations in dotGraph

sle118:
Updating dotGraph to simplity graphical representations by using HTML. Entries and exits are not merged into each node. Sub-states will now be embedded into parent states, which makes it easier to visualize parent-child relationships.


sle118:
Looks like a difference between lf vs crlf broke the build. No time for fixing today. Unless someone else is willing to fix, it'll have to wait

sle118:
Allright, I fixed the build!  Ready for review!

psibernetic:
Can you provide an example for graph output that includes entries and exits and substates? 

Great work on this, I know this will be a huge help to have done for the community. Serious kudos.

src/Stateless/DotGraph.cs
```diff
@@ -1,88 +1,255 @@
...
+                .Add(new Label("placeholder"))
+                .Add(new Style(ShapeStyle.solid));
+            _exitEdges = new FormatList()
+                .Add(new Label("placeholder"))
+                .Add(new Style(ShapeStyle.dotted));
+        }
+
         /// <summary>
-        /// A string representation of the state machine in the DOT graph language.
+        /// A string representation of the stateRepresentation machine in the DOT graph language.
```
psibernetic: This looks like a find replace accident.

src/Stateless/DotGraph.cs
```diff
@@ -1,88 +1,255 @@
...
             }
+        }
+        /// <summary>
+        /// This function builds a representation for nodes and clusters, grouping states together under
+        /// a super state as required.  It is implemented as a recursive that drills through the 
+        /// the sub states when they are found.
+        /// </summary>
+        /// <param name="stateRepresentation"></param>
+        /// <returns></returns>
+        private string BuildNodesRepresentation(StateRepresentation stateRepresentation)
```
psibernetic: I dig it.

nblumhardt:
Thanks for pushing forward with this @sle118! 😎 

I'm curious whether there's a way of factoring this that externalizes more formatting from the `StateMachine` class - I know DOT formatting has been tied up in there for a while, but the footprint (especially the extra fields in _DotGraph.cs_) might be tipping over the healthy limit. Any thoughts on this?

src/Stateless/StateRepresentation.cs
```diff
@@ -276,6 +276,10 @@ public void AddSubstate(StateRepresentation substate)
                 _substates.Add(substate);
             }
 
+            public ICollection<StateRepresentation> GetSubstates()
+            {
+                return _substates;
+            }
             public bool Includes(TState state)
```
nblumhardt: Missing newline.

src/Stateless/FormatList.cs
```diff
@@ -0,0 +1,71 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Stateless.Dot
+{
+    class FormatList
+    {
+        private readonly List<Format> _formats;
```
nblumhardt: The codebase doesn't use explicit `private` modifiers.

src/Stateless/Label.cs
```diff
@@ -0,0 +1,14 @@
...
+{
+    class Label : Format
+    {
+            
+        public Label(string value)
+        {
+            FormatName = "label";
+            FormatValue = value;
+        }
+
```
nblumhardt: Extra whitespace in this file.

src/Stateless/Rank.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        public List<string> EntriesList ;
+        public List<string> ExitsList;
+        public List<string> ClustersList;
+
+        public Rank()
+        {
+            ClustersList = new List<string>();
+        }
+        public static List<string> TryAdd(List<string> list, string entity)
```
nblumhardt: Newline.

src/Stateless/Rank.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System.Collections.Generic;
+
+namespace Stateless.Dot
+{
+    class Rank
+    {
+        public List<string> EntriesList ;
```
nblumhardt: Public fields - could these be properties?

src/Stateless/Rank.cs
```diff
@@ -0,0 +1,35 @@
...
+        public static List<string> TryAdd(List<string> list, string entity)
+        {
+            try
+            {
+                list.Add(entity);
+            }
+            catch
+            {
+                // ignored
+            }
```
nblumhardt: When/why will this throw?

test/Stateless.Tests/DotGraphFixture.cs
```diff
@@ -216,17 +185,30 @@ public void OnExitWithNamedDelegateActionAndDescription()
...
 
             sm.Configure(State.A)
                 .Ignore(Trigger.Y)
                 .Permit(Trigger.X, State.B);
 
             Assert.AreEqual(expected, sm.ToDotGraph());
         }
+
+        [Test]
+        public void test()
```
nblumhardt: Capitalization; may need a more descriptive name :-)

test/Stateless.Tests/DotGraphFixture.cs
```diff
@@ -24,9 +24,8 @@ void OnExit()
         [Test]
         public void SimpleTransition()
         {
-            var expected = "digraph {" + System.Environment.NewLine
```
nblumhardt: Have the expected results changed here, or is it just re-formatting? If the latter, could we revert the re-format so that it's clear things have stayed the same?

src/Stateless/ShapeColor.cs
```diff
@@ -0,0 +1,137 @@
+﻿namespace Stateless.Dot
+{
+    enum ShapeColor
```
nblumhardt: Since these are only used directly in a few places, e.g. literal `ShapeColor.blue`, would it make sense to use a couple of string constants instead?

nblumhardt: Since the same colors could arguably be used for lines and so-on, is `HtmlColor` a better name for the enclosing class?

src/Stateless/Shape.cs
```diff
@@ -0,0 +1,11 @@
+﻿namespace Stateless.Dot
```
nblumhardt: Types in `Stateless.Dot` should be under a `Dot` sub-folder.

test/Stateless.Tests/DotGraphFixture.cs
sle118: Expected results are completely different. I ran each test and validated the output before commit

sle118:
There are a host of improvement that are going to be desirable. For instance the fire async could have some different colour pattern applied, etc

sle118:
@nblumhardt  could you explain what you meant by externalizing?

psibernetic:
I think he means that the State Machine itself should have a clear boundary between functionality and dotGraph code. I would put all of the dotGraph logic in a new class as a strategy pattern for representing a statemachine. You could then make this the default strategy. The StateMachine class does currently violate single responsibility principle.

sle118:
Makes sense. I refractored the original code (from @tlk ?) with modularity in mind. As such, the traversal methods are agnostic of the output format and it would be possible to implement "writers" for different formats.

There is a basic need for some degree of coupling with the state machine, though, as it is easier to manage some elements during state configurations as opposed to downstream implementations which would need to duplicate some of the logic. A good example is description of entries. It is much easier to determine the trigger of an e entry during its definition.

I see there ways to deal with this
- set the description in the stateless classes, which assumes all writers would be limited to a single representation
- define a set of interfaces which a writer would implement, and pass an optional writer to the stateless constructor
- define events to which a writer could subscribe. The writer could be completely separated from the state machine logic. One would attach a writer to the state machine before defining the states (I.e. writer constructor to which a stateless instance is passed) and the writer would collect an internal representation as the state machine is being constructed.

Personally, and if the goal is to reduce the graph footprint, I think the third option would make the most sense.

A reader could also be coded to instantiate the state machine. I like the YAML format, which I successfully used in the past

-

psibernetic:
I'm a fan of #2 personally, as it is simple to implement and solves the immediate problem. With the reader and writer example, I feel like that should be a layer above this library. Deserializing is hard since actions are lambdas. I think the writer in this case should only be concerned with visually or textually representing the machine for information purposes. Thoughts?

psibernetic:
Didn't mean to include an issue link lol.

nblumhardt:
Hope I haven't opened a can of worms :-)

In some sense, we're really talking about the state machine's introspection API here. If you use reflection in .NET as an analogy, one should be able to "reflect over" the state machine instance and see things like states, actions, valid transitions and so-on.

A full introspection API would mean that DOT, plain (`ToString()`) text and other representations could be layered above, as @psibernetic suggests.

Since `ToDotGraph()` and `ToString()` already exist, these would remain but just do something like:

```csharp
public string ToDotGraph()
{
    return new DotGraphFormatter().Format(this);
}
```

Since it doesn't seem like the DOT + HTML format will be accepted by all renderers, would this also be an opportunity to provide separate `DotGraphFormatter` and `DotGraphHtmlFormatter` types?

I'm not sure how much work would ultimately be involved, but it does look like separating the core state machine runtime from the higher-level things like formatting will keep the codebase manageable over time.

nblumhardt:
Hey, just thought I should add - if we think an introspection API is the right way to go, it doesn't necessarily have to be part of this PR. We could put together another PR with the API and then bring this one in on top of it (hopefully only requiring some fairly mechanical transformations).

nblumhardt:
One more data point: a [recent SO question](http://stackoverflow.com/questions/38996425/get-permitted-triggers-with-guard-conditions-in-stateless-net-state-machine-li) that could be answered via an introspection API.

src/Stateless/Rank.cs
sle118: Removed file as code really wasn't needed

sle118:
I have updated the code to reflect the code review comments and pushed the changes.

There is still a lot of room for improving the dotGraph implementation and I am hoping that some can take over from here ( @psibernetic ?) and expand on what I have done. I emphasized mostly on my own use cases, but I'm sure I've missed many of the available patterns.

There is an example of sub-states output in the original thread
https://github.com/dotnet-state-machine/stateless/issues/80#issuecomment-253017141

I like the ideas which have been suggested so far by @nblumhardt  and @psibernetic and it makes perfect sense as long as the proper reflection APIs are implemented in the state machine itself.  For example, I had to implement some defaulting in the StateConfiguration as "OnEntryFrom" were causing some anonymous stuff to show up in the resulting representation.

psibernetic:
I am working on a PR to your branch @sle118 creating the introspection API and placing your code on top of that layer. Working with nested internal classes is SO not fun @nblumhardt. ;)

nblumhardt:
@psibernetic would it be possible to PR in the API first? It might make it easier to discuss/review without taking this thread down the garden path :-)

I would be in favour of extracting the nested classes, if it's possible 👍

nblumhardt:
(Though they do avoid a whole lot of duplicated `<TState, TTrigger>` declarations...)

psibernetic:
I'd love to extract them, but it is a major rework that neither of us has the time for atm I believe.

I will try to pull the API in in a separate PR. I knew a lot was changing so I based on @sle118 branch to get that work, but I don't think they changed much in the normal code. Give me a few to work backwards here.

psibernetic:
@nblumhardt #103 Is ready for review.

sle118:
Does anyone have time to resolve the conflicts after the reflection PR has been merged? I'm swamped at work and won't be able to follow up on this in a while.

nblumhardt:
Four months later.......... #103 is in!

What's the future of this PR now? I am struggling to remember whether any of the functionality from this PR when in via #103, my guess is probably not.

There will definitely be a substantial merge if we go forwards with this branch; it might be easier to grab the  functionality in the PR piecemeal... What do you think, @sle118 ?

betty-crokker:
@sle118 I've merged your changes into my branch, once #158 has been pulled I can quickly get those merges ready with a new PR.  Or, if you want the glory ;-) you can certainly do it!

nblumhardt:
@sle118 closing this now, but would love to have your input as we go forwards (and please feel free to jump back in if you find some time!). Thanks for getting the ball rolling on this!

Continuing via #80 :-)

PR #103, opened 2016-11-15
Reflection API

psibernetic:


psibernetic:
I decided to somewhat eccentric in the naming conventions. ;)

sle118:
Nice work, @psibernetic. So if I understand correctly, the idea is to implement ICartographer in a new cartography class to generate a new format outside of the state machine itself ?

I like that. 

When building my dotGraph class, I hit a couple of limitations. Let's look at the following example:
```
ScanValue = StateMachine.SetTriggerParameters<String>(Triggers.Value);

StateMachine.Configure(State.A)
                .SubstateOf(State.Global)
                .OnEntry(GuiNeedsChangeEvent)
                .OnEntryFrom(ScanValue, OnValidatingValue)
                .OnEntryFrom(Triggers.DefaultVal, OnValidatingValue)
                .Permit(Triggers.OK, State.NextState);
``` 

OnEntryFrom gets a trigger with parameters and creates a new entry action from it. The issue is that the triggers with parameters don't quite look nice using reflection and this makes it difficult to properly represent the transition by default.

So ideally, I think the StateResource class should be responsible to normalize the representation of a trigger  to simplify the traversal of Entry Actions. 

I'm thinking about some possible approaches:
- it could expose "Trigger" (the raw object), alongside "UnderlyingTrigger" which would be either a simple Trigger or the underlying trigger of a triggerWithParameters. 
- it could extract a tidy normalized description of the underlying trigger as well as name of parameters if that's also possible (or their types).

Is this the best place to discuss this, or should this be in the original thread?

psibernetic:
Thanks, and yes, this is my take on a combination of Visitor pattern and [IFormatProvider](https://msdn.microsoft.com/en-us/library/system.iformatprovider(v=vs.110).aspx).

This is probably the right place to discuss what we want to expose in the StateResource objects for first pass, keeping in mind that adding things is easy, removing or changing will be breaking changes for implementors. (Though the IExplorer interface could handle versioning of StateResource object graphs).

I do think the right approach is to expose triggers as an object with a "TypeParameters" collection. In general any niceties or normalization should happen in a cartographer or extension method. I'll add an object for that tonight when I get home. Glad you like the idea.

nblumhardt:
Nice :-)

I think the name's cool, but I wonder if modelling on .NET reflection could remove the need for the interface/visitor bit?

```csharp
var sm = new StateMachine<State,Trigger>();

// Analogous to System.Type and object.GetType(), note
// it's non-generic:
StateMachineInfo info = sm.GetInfo();

// StateType is a System.Reflection.TypeInfo; returns typeof(TState)
Console.WriteLine(info.StateType);

// *Info types like StateInfo analogous to MemberInfo, ParameterInfo etc.
foreach (StateInfo state in info.States)
{
    // The property would be of type object, but return
    // a boxed TState object
    Console.WriteLine(state.UnderlyingState);
}
```

With a `StateMachineInfo` at the root, it would mean the API for things like formatters might be:

```csharp
var formatter = new DotGraphFormatter();

// Because StateMachineInfo is non-generic, DotGraphFormatter
// and Write() don't have to be generic.
formatter.Write(sm.GetInfo(), ...);
```

Just as with reflection, a number of scenarios could be implemented on top, e.g. analysis, "convention tests" and so-on - we wouldn't be coupling the API to the idea of mapping out or printing the graph.

Maybe the mapping to reflection APIs is a bit of a stretch - but I like the idea that _"if you know how to use reflection, you know how to reflect over StateMachine"._ Any ideas?

psibernetic:
The Explore method does expose the internal structure via the StateResource structure if you wanted to use it for some other purpose. The Cartographer in-fact uses that method to do its discovery. The IExplorer interface and StateResourceExplorer are really just ways to move the logic of mapping layer between StateRepresentations to StateResources to another class as separation of concerns. 

Regarding using reflection APIs, I suppose we could include that information if we want, but the textual representation of TState does depend on the instance of type TState, we could expose StateInstance as an object or TState directly, but in most cases the ToString() is what we want, no?

I'm totally open to merging this to a new branch in this repo for collaboration until we get the StateResource definition squared up.

sle118:
Take a minute to read this thread and think about the level of abstraction this has reached between total strangers that are only connected through some collective creation. Having such conversations over a beer at the pub would result in side looks from the people around wondering what sort of wizardry is going on. This place is great!

nblumhardt:
> The Explore method does expose the internal structure via the StateResource structure if you wanted to use it for some other purpose.

@psibernetic 👍 I should have been a bit clearer, sorry - the snippet I posted pretty much _is_ `Explore()`, `StateRepresentation` and so-on, just re-cast in the reflection analogy.

> in most cases the ToString() is what we want, no?

For display, yes, but for (a hypothetical) example, imagine writing a unit test that checks to make sure every member in an `enum` is represented as a state - in those kinds of cases, it'll be nicer to work with the real values. Even for display, you could imagine writing a visualizer for a state machine that used additional information from the state type to provide more information.

I'm happy to kick off a branch and dig into it there, though we might miss the nice diff/comment/discussion view that the pull request format brings. If it helps work around time constraints that's probably a worthwhile trade-off though, is that what you have in mind?

psibernetic:
@nblumhardt I do see your point, and agree for the most point, i'll hit some key points tonight. 

Once I make my changes, we could merge into a branch here (cartographer-api). From there we can open a new PR from that branch to dev to track the WIP. Each of us can make PR to the dotnet-state-machine/stateless/cartographer-api branch to collaborate. May end up helping discussion by breaking it into actionable items.

psibernetic:
Not getting to this tonight, work got in the way, try again tomorrow.

sle118:
I wouldn't want to send us into a rabbit hole, but how would you call a reverse cartography API?  Using the cartography API, one could persist a state machine to disk. Using a "builder" (can't think of anything better now), these could be loaded from disk.

From there, maybe someone would write a graphical editor similar to this one below
https://github.com/vosmont/Vera-Plugin-RulesEngine

That editor would bee called in visual studio akin to the visual studio xaml designer. 

The same persist/load mechanism would also help long running workflows that span across process restart or that allow reducing the memory footprint when dealing with parallel processing.

I certainly don't want to overblow this tidy project, though. I'm just throwing ideas to see if they will stick

Let me know your thoughts!

psibernetic:
@sle118 That definitely is out of scope for this project. StateMachines here will inherently be used in-proc and not persisted. 

I am working on a StateMachine-As-A-Service project that can do pretty much what you are talking about. It is a work-in-progress, but the service actually works, just needs more connectors and a UI. I can work with you on that project if you want, just ping me over there: https://github.com/psibr/REstate

HenningNT:
@sle118 At work we use the [Quantum Leap state machine framework](http://www.state-machine.com/)  for the embedded code. We use a graphical editor to design and generate the state machine.
 I've been considering making a tool of some sorts to import the model file (XML-ish) or output of this state machine diagram editor in some way, either directly at runtime, or just a code generator. I've also been thinking of creating something that will create model files so I can view those in QM.

sle118:
I knew this was a rabbit hole.  I read about Quantum Leap but never used it.  I do a lot of small home automation projects using the "MySensors" framework on Arduino but didn't have a project that justified the learning curve of the Quantum Leap Tool. I'm not sure what sort of optimisation there is for a processor like the atmega 328. 

@psibernetic You might have provided an answer for a question that has been lingering in the back of my mind for quite some time now. I'm working on a small cellar management software (which has limited GUI and mostly interacts with the users using a combo of voice feedback and barcode reader encoded triggers) which heavily uses the stateless state machine as its user interaction backbone. 

The state machine is the core engine which defines all possible actions given the current context, and the gui merely passes triggers down and subscribes to "update events". So far it's been all great, but I am scratching my head over publishing the core "cellar" engine as a restful service (therefore stateless) so I'll definitely have a look. My bandwidth is limited; projects had wound down at work, but the quiet time is over now and I'll have to hold on to the personal stuff for a little while. I'll get to it at some point, though.

nblumhardt:
Just FWIW :-) one of the reasons `StateMachine` is generic is that `TState` and `TTrigger` can also just be strings.

Loading from file:

```csharp
var sm = new StateMachine<string, string>("initial");

// E.g. on each line a state, trigger, destination tuple
foreach (var transition in LoadStateTriggerDestinations("file.csv"))
{
    sm.Configure(pair.State).Permit(pair.Trigger, pair.Destination);
}

sm.Fire("Go");
```

sle118:
@nblumhardt  my brain is on fire! ;)

psibernetic:
I've updated this to include type'd states/trigger where necessary or nice, didn't change anything else, I think I'll leave that to you to decide on @nblumhardt.

nblumhardt:
Thanks Ovan, will jump into this as soon as I have a chance 👍

nblumhardt:
Hey folks, happy 2017! :-) Hope you had a good break @psibernetic.

Returning to this now, I very much want to move this forward, but I am unsure about the design (reflection-like vs. visitor pattern) and think we'll just create more work for ourselves if we merge into a new branch as-is and then try and massage it into a different shape.

With some time to think about it, I'm still keen to head more in the direction of https://github.com/dotnet-state-machine/stateless/pull/103#issuecomment-260783635. If we went that direction, `StateResource` from the PR would be fairly close to `StateInfo`, and the code that builds it would be practically the same, but the organisation would be quite different. What's your current thinking, Ovan?

Cheers!
Nick

betty-crokker:
I'm not sure what the etiquette is for jumping in on a thread like this <grin> but I just downloaded Stateless and am excited about what it can do for me!  And being able to generate DOT files that have all the substates indicated, will make it all the sweeter.  
I agree with you, Nick, I would prefer reflection to a visitor pattern, in my experience it's easier to use and it's more common in C#.  Can I be any help in moving this feature forward?

Chris

nblumhardt:
Hi @betty-crokker - thanks for jumping in!

@psibernetic what are your thoughts? Cheers!

psibernetic:
Hey @nblumhardt, @abalfour did some work to this PR for us. Time to review that. Also, this merge conflict isn't quite as straight forward since that file no longer does the work of mapping to DOTGraph, can you make sure we aren't regressing in the DotGraphCartographer?

nblumhardt:
@psibernetic @abalfour the changes are definitely heading in the right direction 👍 

I think there's still a lot of ground to cover; if we're heading towards https://github.com/dotnet-state-machine/stateless/pull/103#issuecomment-260783635 there are some more changes that should be pretty mechanical to make (non-generic info type, some class names and method layout).

Is the plan for @abalfour to continue working on the PR? To get this landed, I think the next step would be to get this to a final cut so we can do one really solid review pass.

We can probably circle back on testing `DotGraphCartographer` (-> `DotGraphFormatter`?) when the rest of the API is settled.

nblumhardt:
Looks awesome, thanks @psibernetic!

I will run through and see if I can spot any issues, with an eye to getting this merged. Looking great overall 👍

nblumhardt:
Okay, had a bit more of an explore. Mostly just a list of nitpicks - I can PR these to your branch @psibernetic if that suits? (Or just directly edit if that'd be easier.)

 * The `Info` classes would benefit from being immutable, e.g. take the various elements as ctor parameters and expose read-only properties (this'd allow for things like caching perhaps in the future)
 * I don't think we need to declare `IStateMachineInfoFormatter` just yet; aside from encouraging consistency (which is good) I don't think there will be a lot of points where it's used, so we could hold off and not tie ourselves down to a new type here at this stage
 * I think we should rev the major version and get rid of `ToDotGraph()`, since it's easy to document `new DotGraphFormatter().Format(sm.GetStateMachineInfo())`, and this will make alternative formats (like HTML DOT discussed in the other PR) on an equal footing
 * `GetStateMachineInfo()` could just be `GetInfo()`, since it's being invoked on a `StateMachine`
 * `StateBindingInfo` could just be `StateInfo` (we don't use the term "binding" elsewhere at this stage)

Thanks again for driving this one, it's a great PR 👍

SaphuA:
Wow this is exactly what I was looking for. Any idea when this will be production ready? Do you have a list of todos? Perhaps I could do a little work for you.

psibernetic:
I think this is all that that is left.

- Remove IStateMachineInfoFormatter interface.
- Remove ToDotGraph() from StateMachine class.
- Remove Set; from all Info classes.
- Expose all properties that had Set; removed in a constructor instead. (Serializability effort).
- Add unit test for StateMachine -> StateMachineInfo conversion.

If anyone wanted to make a PR to my branch, that would be fine. I'm working on this as I have time, which is sparse at the moment. Thanks for the interest guys!

betty-crokker:
I'm familiar with git, but not the particular ways it's used here on github.

I made a branch:

https://github.com/betty-crokker/stateless/tree/cartographer-api

and started the work:
* IStateMachineInfoFormatter class removed
* ToDotGraph() removed from StateMachine
* Property setters marked "internal" or "private" for Info classes
* Created a new unit test class BindingInfoFixture, for testing the new StateMachineInfo.

Am I doing this right?  Should I do the "Pull Request" when I'm done making changes, or now, or should I have done that instead of just making a branch?

I took the existing SimpleTransition test, renamed it SimpleTransition_DotGraph, and copied it into my new BindingInfoFixture test class and named it SimpleTransition_Binding.  Sound good?  My thought is to take each test in DotGraphFixture and make a parallel in BindingInfoFixture that has the same state machine.

Can you take a look at my new SimpleTransition_Binding unit test, see if you like the style?  I'm not a big fan of 'var'.  If that first unit test looks good I'll keep writing them.

psibernetic:
@betty-crokker Lets do this:

Create a PR to my branch, changes will show up here as they merge in. I will give you contributor access to my fork to make that easier, so you can merge your PR work without me around.

betty-crokker:
I think the reflection classes are pretty much ready for prime time, it would be good to exercise them a little by actually doing some of the updates to the DotGraphFormatter.  Had anyone started work on that, or is that where we can next apply some brain power?

nblumhardt:
This has been an epic PR :-)

There are still some items from https://github.com/dotnet-state-machine/stateless/pull/103#issuecomment-281199027 I want to look at, but it's now probably more efficient for everyone to collaborate on `dev`. I'll merge this shortly, and we can pick up the remainder of the DotGraph stuff and API nitpicks there.

Thanks for pushing things along @betty-crokker, and thanks for the huge effort on this @psibernetic!

nblumhardt:
Out of time this afternoon, I've done what I can and created #143 with the remaining TODOs I can think of; any help over there would be greatly appreciated. Cheers all! :-)

PR #106, opened 2016-11-22
Removed unneeded dependency NETStandard.Library for NET4x

Lakritzator:
As all the code is already in the framework (.NET 4.x is NETStandard 1.0 compatible) the dependency adds a lot of noise to .NET Framework projects. I didn't detect this right away, as System.Reactive made the same "mistake" but now corrected it with 3.1.1

See https://github.com/Reactive-Extensions/Rx.NET/commit/5f3ff8deb1ee6fa4434c76cb12fc91b3aa0cd426
for a similar change.

So here is a pull-request which removes the unneeded dependency.

This change didn't cause issues with dotnet restore / build, more tests were from my standpoint not needed as in the end the dependencies are not copied to the target directory (for NET4x).

I would event suggest to place this into master, and release a hotfix like System.Reactive did..

Keep up the good work!!!!

### DIFF
--- a/src/Stateless/project.json
+++ b/src/Stateless/project.json
@@ -1,7 +1,6 @@
 {
   "version": "3.0.1-*",
   "dependencies": {
-    "NETStandard.Library": "1.6.0"
   },
   "packOptions": {
     "licenseUrl": "http://www.apache.org/licenses/LICENSE-2.0",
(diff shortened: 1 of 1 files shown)


Lakritzator:
Better now? :)
And no, it shouldn't matter 1.6.0 or 1.6.1: https://blogs.msdn.microsoft.com/dotnet/2016/11/16/announcing-net-core-1-1/

HenningNT:
Looks good to me, lakris!

nblumhardt:
Looks good, thanks 👍 - if anyone can smoke test the resulting dev package, I agree we should push this through to master as well. Cheers!

Lakritzator:
3.0.1-dev-00123 solved my dependency issues, everything seems to work but I didn't do any thorough tests on the state engine itself. So from my point of view #107 can be merged,

nblumhardt:
Thanks. There are no changes to logic in that version, so successful package installation is probably all we need to look out for. Merging, thanks!

PR #112, opened 2016-12-03
Multiple guards support

edgars-pivovarenoks:
Hi, I propose to support multiple transition guards as I had this requirement in real-world project. 

My idea is to replace pair of guard and description parameters with TransitionGuard class and use TransitionGuards class internally to hold guard functions and return aggregate results. Below is configuration code fragment : 

```
 .PermitIf(Trigger.Close, State.Closed, g => g
     .And(() => true, "my guard 1")
     .And(() => true, "my guard 2"))
```
Current implementation requires all conditions to be true to allow transition.

Will be glad to hear feedback about this idea and input about better implementation.


HenningNT:
Would you also add `Or(() => true, "my guard 3"))` ;-)

But can't you just call a method that evaluates several conditions ?

Other than that I think it sounds like a good idea.

edgarspiv:
You are right about method with several conditions, that is exactly how I solved it so far. What I was not able to solve (and forgot to mention in) is when such guard fails, we have have fixed `InvalidOperationException` and we don't know which conditions failed.

So this was my primary motivation - to get finer grained report of failed conditions, making it easier to find bugs and have better error reporting to users.

Perhaps, another alternative is to let override `DefaultUnhandledTriggerAction` leave this up to implementer, but I have not investigated this option enough whether it's feasable.

nblumhardt:
Thanks for the PR @edgars-pivovarenoks ! I like the improved error reporting, really makes sense to expose all the information we can.

I'm not against the lambda-based configuration API, but just to play devil's advocate, I sometimes find these kinds of interfaces a bit harder to discover via IntelliSense than other options.

Just throwing some ideas out there, since we can expect tuple language support in C# 7 could we consider doing a `params Tuple<Action<bool>, string>` version like:

```csharp
 .PermitIf(Trigger.Close, State.Closed,
    (() => true, "my guard 1"),
    (() => true, "my guard 2"))
```

(First thing that came into my head - could be some other options I'm not considering.)

Impressions...?

edgars-pivovarenoks:
@nblumhardt, regarding configuration API, agreed. 

> could be some other options I'm not considering

I had this thought as well, like I am missing something, but decided to give it a try. I don't have other ideas at the moment, so will work on it, then see how it looks.

nblumhardt:
Sounds good, thanks Edgars!

src/Stateless/StateRepresentation.cs
```diff
@@ -57,6 +57,7 @@ bool TryFindLocalHandler(TTrigger trigger, out TriggerBehaviour handler, params
                     return false;
                 }
 
+                // TODO : Refactor for condition results
```
nblumhardt: Does anything need to happen here?

edgars-pivovarenoks: That was WIP, latest commit contains updated TryFindLocalHandler().

edgars-pivovarenoks:
I pushed pretty much what I had on mind. 

Little summary of last changes :
- new class TriggerBehaviourResult, short lived result holding list of failed guard conditions,
- Refactored TryFindHandler to be called only **once per Fire** in case of failed guard,
- UnhandledTriggerAction got new parameter with list of unmet condition,
- Error message updated in default unhandled trigger action.

src/Stateless/StateMachine.cs
```diff
@@ -283,7 +283,7 @@ void InternalFireOne(TTrigger trigger, params object[] args)
         /// is fired.
         /// </summary>
         /// <param name="unhandledTriggerAction">An action to call when an unhandled trigger is fired.</param>
-        public void OnUnhandledTrigger(Action<TState, TTrigger> unhandledTriggerAction)
+        public void OnUnhandledTrigger(Action<TState, TTrigger, ICollection<string>> unhandledTriggerAction)
```
nblumhardt: Could this be added as an overload to avoid the breaking change here?

src/Stateless/StateRepresentation.cs
```diff
@@ -38,45 +38,50 @@ public StateRepresentation(TState state)
...
+                // Guard functions executed
+                var actual = possible
+                    .Select(h => new TriggerBehaviourResult(h, h.UnmetGuardConditions));
+        
+                handlerResult = TryFindLocalHandlerResult(trigger, actual, r => !r.UnmetUnmetGuardConditions.Any())
+                    ?? TryFindLocalHandlerResult(trigger, actual, r => r.UnmetUnmetGuardConditions.Any());
+
+                return !handlerResult.UnmetUnmetGuardConditions.Any();
+            }
+            TriggerBehaviourResult TryFindLocalHandlerResult(TTrigger trigger, IEnumerable<TriggerBehaviourResult> results, Func<TriggerBehaviourResult, bool> filter)
```
nblumhardt: Nit: missing newline above.

nblumhardt:
Thanks for the follow-up Edgars!

I'm a bit unclear on the need for the additional public classes - if the guard condition can be configured as a tuple of predicate and string, do we need all of the other types on the public interface?

src/Stateless/StateMachine.cs
edgars-pivovarenoks: Absolutely.

edgars-pivovarenoks:
Nicholas thanks for feedback, will implement those fixes shortly. 

Regarding class access level, thanks for noticing. 

To clarify, what's your opinion should we keep TransitionGuard, GuardCondition classes at all? 

My motivation to not use tuple beyond public API is for stake of readability (due to tuple member names). Similar with TransitionGuard class, at this point is more to communicate the intent, specifically in TriggerBehaviour class.

edgars-pivovarenoks:
Pushed updates, if you see something, let me know.

Meanwhile I will :
- think what tests could be added,
- check if DOT graph generation looks good.

HenningNT:
So if I understand it correctly we can do like this:
```_machine.Configure(State.Open)
     .PermitIf(Trigger.Assign, State.Assigned,
       new Tuple<Func<bool>, string>[]
         { new Tuple<Func<bool>, string>(GuardIsOnVacation, "The boss is on vacation"),
           new Tuple<Func<bool>, string>(GuardNotOnFriday, "No bug state change allowed on Fridays")} );
```

Can I suggest we might want PermitReentryIf with multiple guards as well:

            public StateConfiguration PermitReentryIf(TTrigger trigger, params Tuple<Func<bool>, string>[] guards)
            {
                return InternalPermitIf(
                    trigger,
                    _representation.UnderlyingState,
                    new TransitionGuard(guards));
            }

edgars-pivovarenoks:
Henning, little simpler with the current C# because we have condition array as _params_
```
.PermitIf(Trigger.Assign, State.Assigned,
         new Tuple<Func<bool>, string>(GuardIsOnVacation, "The boss is on vacation"),
         new Tuple<Func<bool>, string>(GuardNotOnFriday, "No bug state change allowed on Fridays"));

```

of course, it could not compete with how laconic it would be in C# 7 :
```
 .PermitIf(Trigger.Close, State.Closed,
    (GuardIsOnVacation, "The boss is on vacation"),
    (GuardNotOnFriday, "No bug state change allowed on Fridays"));
```
I really like that with C#7 it looks just as a small augmentation to existing syntax, very easy refactoring if necessary.
```
 .PermitIf(Trigger.Close, State.Closed,
    GuardIsOnVacation, "The boss is on vacation");
```

Regarding other methods like PermitReentryIf, 
indeed, it appear logical to provide same API for other methods that currently have guard parameters. To be honest, I have postponed it until after it is settled whether this implementation goes forward and you guys like it.

src/Stateless/TransitionGuards.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Stateless
+{
+    public partial class StateMachine<TState, TTrigger>
+    {
+        internal class GuardCondition
```
nblumhardt: Nit: should be in _GuardCondition.cs_

src/Stateless/TransitionGuards.cs
```diff
@@ -0,0 +1,37 @@
...
+            internal GuardCondition(Func<bool> guard, string description)
+            {
+                _guard = guard;
+                _description = description;
+            }
+            internal Func<bool> Guard { get { return _guard; } }
+            internal string Description { get { return _description ?? _guard.TryGetMethodName(); } }
+        }
+
+        internal class TransitionGuard
```
nblumhardt: Nit: should be in _TransitionGuard.cs_

src/Stateless/StateRepresentation.cs
```diff
@@ -38,45 +38,51 @@ public StateRepresentation(TState state)
 
             public bool CanHandle(TTrigger trigger)
             {
-                TriggerBehaviour unused;
+                TriggerBehaviourResult unused;
                 return TryFindHandler(trigger, out unused);
             }
 
-            public bool TryFindHandler(TTrigger trigger, out TriggerBehaviour handler)
+            public bool TryFindHandler(TTrigger trigger, out TriggerBehaviourResult handler)
```
nblumhardt: The tricky cases here around substates with unmet conditions, superstates with unmet conditions and so-on seem to warrant some tests - it would be really useful to illustrate these scenarios so that they're not inadvertently broken by future updates.

src/Stateless/StateMachine.Async.cs
```diff
@@ -144,7 +144,7 @@ public Task FireAsync<TArg0>(TriggerWithParameters<TArg0> trigger, TArg0 arg0)
         /// is fired.
         /// </summary>
         /// <param name="unhandledTriggerAction">An asynchronous action to call when an unhandled trigger is fired.</param>
-        public void OnUnhandledTriggerAsync(Func<TState, TTrigger, Task> unhandledTriggerAction)
+        public void OnUnhandledTriggerAsync(Func<TState, TTrigger, ICollection<string>, Task> unhandledTriggerAction)
```
nblumhardt: The old `OnUnhandledTriggerAsync()` overload is missing here (for the non-async version we ended up with a new overload alongside the old one)

nblumhardt:
Tests are looking good, thanks for pushing forward with it! I need to spend a little more time working through the test cases, but apart from the missing `OnUnhandledTriggerAsync()` overload I think we're good to go! :-)

src/Stateless/StateMachine.Async.cs
edgars-pivovarenoks: Async, right.

edgars-pivovarenoks:
Any objections if I add overloads for other methods that currently have guards? @HenningNT mentioned PermitReentryIf() for example.

nblumhardt:
Ah good point! Yes, it would make sense to be consistent there I think.

edgars-pivovarenoks:
List of methods now having multiple condition overloads :
- PermitIf()
- PermitReentryIf()
- IgnoreIf()
- PermitDynamicIf()
- PermitDynamicIf<TArg0>()
- PermitDynamicIf<TArg0, TArg1>()
- PermitDynamicIf<TArg0, TArg1, TArg2>()

nblumhardt:
Awesome :-) I'm holidaying this week, but should be able to get this merged early next. Thanks for all the work on this!

nblumhardt:
:+1:

PR #114, opened 2016-12-05
Fixed dialed/dialled typo in README.md

matthiaslischka:
There where both versions of dialed/dialled in the README.md

I know its just a typo and a PR is probably a overhead but I thought why not fix it right away instead of creating an issue or ignoring it.

For an explanation on the correct writing of dialing/dialling see:
http://www.future-perfect.co.uk/grammar-tip/is-it-dialled-or-dialed/
I went for the british english version. However, if you decide to go for the american version, just make sure it is straight throughout the readme.md as there is a mix right now. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 var phoneCall = new StateMachine<State, Trigger>(State.OffHook);
 
 phoneCall.Configure(State.OffHook)
-    .Permit(Trigger.CallDialed, State.Ringing);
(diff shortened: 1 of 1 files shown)


nblumhardt:
Thanks Matthias :-) - I see there's a "behaviour" in there as well (and I think I still wrote READMEs using Australian English back when I started this project), so the British spelling seems like a fair choice 👍

Issue #116, opened 2016-12-07
PCL Target


fvaillancourt:
Hi, we are working on a multi-platform project which use a lot of PCL projects. Also, as we are using a lot of async code, we would like to use version 3 of Stateless rather than 2.5.x. However it seems the PCL target was removed in version 3. We are aware that the world is moving away from PCL but it is not possible to switch all the code base in a day. Could you add the ".NETPortable,Version=4.5,Profile=Profile7" target or the same target that was supported by version 2.5.x to help ease the transition?
I tested it by adding only the target below in project.json and everything seemed to compile and work fine.
```
    ".NETPortable,Version=4.5,Profile=Profile7": {
      "buildOptions": {
        "define": [ "PORTABLE_REFLECTION", "TASKS" ]
      }
```

nblumhardt:
Hi @fvaillancourt - thanks for getting in touch. I think we can do this (unless anyone else can see a downside?), would you be interested in submitting a PR? Thanks!

fvaillancourt:
The -dev-00132 was successfully referenced by a PCL project, itself referenced by a Xamarin iOS app.

nblumhardt:
Great, thank you.

PR #117, opened 2016-12-08
Added PCL target profile 259

fvaillancourt:
Added PCL target profile 259 like Stateless 2.5.x so that projects that are not completely caught up with .NETStandard can still use Stateless 3.x

### DIFF
--- a/src/Stateless/project.json
+++ b/src/Stateless/project.json
@@ -26,9 +26,17 @@
     },
     "net4.5": {
       "buildOptions": {
-        "define": ["TASKS"]
+        "define": [ "TASKS" ]
       }
     },
-    "net4.0": {}
+    "net4.0": {},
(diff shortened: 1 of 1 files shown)


fvaillancourt:
This is my first pull request, I hope I did it right. Also I do not have that much .NET Standard experience so let me know if it is not right.

nblumhardt:
Wonderful! Thanks for taking the time to submit this. Hopefully it's the first of many PRs for you, glad we could get you started :-)

This is in reference to #116 (if you add a `#` then an issue number, GitHub will link the issues).

I'll merge this, after which a `-dev-nnnnn` package will be automatically published to NuGet. If you can smoke-test it by installing it with the `-Pre` flag, and let us know the results, we'll eventually merge it over to `master` and the released version of the package.

Cheers!

Issue #121, opened 2017-01-02
Accidental configuration of a recursive substate definition is allowed and will result in a stack overflow


smartcaveman:
The accidental configuration of a recursive substate definition is allowed and will result in a stack overflow.  It happens in 

It seems like a better design would raise more useful and immediate error.   
- One way to do this is to validate the `StateConfiguration<TState, TTrigger>.SubstateOf(TState) ` input and raise an error if the definition is recursive.
- Another approach would be to address this where the error occurs in `StateRepresentation<TState,TTrigger>.Includes` by adding a second condition to the `Any()` predicate.   

I would recommend the first approach because it seems like defining a state as a substate of itself should be recognized as an error.

HenningNT:
I agree to adding a check at the configuration phase. I think parameter validating should be done in the publicly accessible methods. After the parameters have been checked the state machine configuration should be runnable.

HenningNT:
I think this is quite important to get in, so I submitted a new PR to cover this, and a bit more: #124

smartcaveman:
Great!  Thanks for pushing this through and resolving the issue.

HenningNT:
I think we can close this one.

nblumhardt:
:+1:

PR #122, opened 2017-01-02
Added validation to `SubstateOf`

smartcaveman:
Added validation against recursive substate definitions in SubstateOf.  Details of problem @ https://github.com/dotnet-state-machine/stateless/issues/121

### DIFF
--- a/src/Stateless/StateConfiguration.cs
+++ b/src/Stateless/StateConfiguration.cs
@@ -453,6 +453,9 @@ namespace Stateless
             /// <returns>The receiver.</returns>
             public StateConfiguration SubstateOf(TState superstate)
             {
+                if(_representation.UnderlyingState.Equals(superstate)){ 
+                    throw new InvalidOperationException(string.Format("A given state cannot be defined as a substate of itself.  The problem state is {0}.", superstate));
+                }
                 var superRepresentation = _lookup(superstate);
                 _representation.Superstate = superRepresentation;
                 superRepresentation.AddSubstate(_representation);


nblumhardt:
I like this in principle, but it seems inconsistent that we'd detect `S : S`, but not `S : S' : S`, i.e. cycles involving more than one state.

Would it be possible to detect the issue at the point the overflow is triggered (guessing it is during `Fire()`) and reporting there?

HenningNT:
Although the effort of checking if the superstate is valid or not is low, I don't think it needs to be done every time Fire is executed. I think we should assume that if the state machine completes the configuration step without errors, then the configuration should be correct. So we check every parameter the user provides at the public methods...

This modified SubstateOf method checks for identical and nested identical superstates.

            public StateConfiguration SubstateOf(TState superstate)
            {
                // Check for accidental identical cyclic configuration
                if (State.Equals(superstate)) { throw new ArgumentException($"Configuring {State} as a substate of {superstate} creates an illegal cyclic configuration!"); }

                // Check for accidental identical nested cyclic configuration
                List<TState> superstates = new List<TState>();
                superstates.Add(State);

                // Build list of super states
                var activeRepresentation = _lookup(superstate);
                while (activeRepresentation.Superstate != null)
                {
                    superstates.Add(activeRepresentation.Superstate.UnderlyingState);
                    activeRepresentation = _lookup(activeRepresentation.Superstate.UnderlyingState);
                }
                if (superstates.GroupBy(n => n).Any(c => c.Count() > 1))
                { throw new ArgumentException($"Configuring {State} as a substate of {superstate} creates an illegal nested cyclic configuration!"); }

                // All checks passed, it's safe to create this configuration
                var superRepresentation = _lookup(superstate);
                _representation.Superstate = superRepresentation;
                superRepresentation.AddSubstate(_representation);
                return this;
            }

nblumhardt:
@HenningNT aha! Good thinking 👍

smartcaveman:
Looks good.  I'll mention one minor possible optimization that I notice: `if (superstates.GroupBy(n => n).Any(c => c.Count() > 1))` would be better as `if (superstates.GroupBy(n => n).Any(c => c.Any()))`.  I don't think it's worth changing, b/c it wouldn't make a real difference unless someone is misusing the library, but if for some reason there were a really (impractically) huge number of states it could have an impact.

Thanks for fixing this!

smartcaveman:
Thanks!

nblumhardt:
@smartcaveman I don't think anything's fixed - as I understand it, @HenningNT is just suggesting a way it could be approached (unless I misunderstood). Cheers!

HenningNT:
Sorry, I didn't intend to hijack the PR, I just wanted to suggest a way to solve the problem.
If you want to, I can submit a new PR, it's up to you.

Cheers!

PR #124, opened 2017-01-13
Adding check for Cyclic configuration

HenningNT:
I'm moving the checks to the Enforce class for consistency.
I added NotIdentityTransition and NotCyclicConfig to the Enforce class.
I also replaced code like 'if (entryAction == null) throw new ArgumentNullException(nameof(entryAction)); ' with calls to Enforce.ArgumentNotNull so that the checks are done the same way.


src/Stateless/StateConfiguration.Async.cs
```diff
@@ -21,7 +21,7 @@ public partial class StateConfiguration
             /// <returns></returns>
             public StateConfiguration InternalTransitionAsync(TTrigger trigger, Func<Transition, Task> entryAction)
             {
-                if (entryAction == null) throw new ArgumentNullException(nameof(entryAction));
+                Enforce.ArgumentNotNull(entryAction, nameof(entryAction));
```
nblumhardt: Given the built-in support for the `if (arg == null) throw ...` pattern in tools like ReSharper, it might be time to retire the `Enforce` pattern used in this lib. What do you think?

src/Stateless/Enforce.cs
```diff
@@ -15,5 +14,35 @@ public static T ArgumentNotNull<T>(T argument, string description)
...
+                throw new ArgumentException(StateConfigurationResources.SelfTransitionsEitherIgnoredOrReentrant);
+            }
+        }
+
+        public static void NotCyclicConfig<TState, TTrigger>(StateMachine<TState, TTrigger>.StateRepresentation _representation, TState superState, Func<TState, StateMachine<TState, TTrigger>.StateRepresentation> _lookup)
+        {
+            var State = _representation.UnderlyingState;
+
+            // Check for accidental identical cyclic configuration
+            if (State.Equals(superState)) { throw new ArgumentException($"Configuring {State} as a substate of {superState} creates an illegal cyclic configuration!"); }
```
nblumhardt: Generally exceptions end with a full stop `.` rather than exclamation mark `!`.

src/Stateless/Enforce.cs
```diff
@@ -15,5 +14,35 @@ public static T ArgumentNotNull<T>(T argument, string description)
 
             return argument;
         }
+        public static void NotIdentityTransition<TState, TTrigger>(StateMachine<TState, TTrigger>.StateRepresentation _representation, TState destination)
```
nblumhardt: I'd rather not go down the path of moving all validation to `Enforce` - it reads nicely at the call site, but it'll lead to a jumble of unrelated things here. `StateConfiguration` still seems like the right home for these.

src/Stateless/StateConfiguration.Async.cs
HenningNT: I don't mind!
Actually when I first started using Stateless, I thought that the Enforce class was a bit unnecessary.
I'll remove the references to Enforce.

src/Stateless/Enforce.cs
HenningNT: ok

src/Stateless/Enforce.cs
HenningNT: Yes, I see, it's even recommended not to use exclamation marks.

HenningNT:
Hmmm, I don't think I should have done the last committ... Is there a way to undo it?

nblumhardt:
Hi Henning,

What you'll need to do depends on your local working copy's state. If you haven't made any subsequent commits:

```
git reset --hard HEAD~1
git push origin dev --force
```

_should_ do the trick. Some fiddling may be necessary :-)

Make sure your local copy looks exactly as you want it before the force-push :-)

HenningNT:
Ok, thanks, I'll give that a go!

PR #125, opened 2017-01-19
Adding check for cyclic configuration

HenningNT:
Not sure what happened to the previous PR, it got canceled automatically when I did the repository undo...
Anyway, here's a new fresh PR, which changes one thing only.
I added two tests to verify the the check throws ArgumentException, the reviewer might want to double check the test to verify that the created state machine is actually a cyclic configuration :-D

### DIFF
--- a/src/Stateless/StateConfiguration.cs
+++ b/src/Stateless/StateConfiguration.cs
@@ -511,23 +511,27 @@ namespace Stateless
                 var State = _representation.UnderlyingState;
 
                 // Check for accidental identical cyclic configuration 
-                if (State.Equals(superstate)) { throw new ArgumentException($"Configuring {State} as a substate of {superstate} creates an illegal cyclic configuration."); }
+                if (State.Equals(superstate))
+                {
+                    throw new ArgumentException($"Configuring {State} as a substate of {superstate} creates an illegal cyclic configuration.");
+                }
 
                 // Check for accidental identical nested cyclic configuration 
-                List<TState> superstates = new List<TState>();
-                superstates.Add(State);
+                var superstates = new HashSet<TState> { State };
 
-                // Build list of super states 
+                // Build list of super states and check for 
                 var activeRepresentation = _lookup(superstate);
                 while (activeRepresentation.Superstate != null)
                 {
+                    // Check if superstate is already added to hashset
+                    if (superstates.Contains(activeRepresentation.Superstate.UnderlyingState))
+                        throw new ArgumentException($"Configuring {State} as a substate of {superstate} creates an illegal nested cyclic configuration.");
+
                     superstates.Add(activeRepresentation.Superstate.UnderlyingState);
                     activeRepresentation = _lookup(activeRepresentation.Superstate.UnderlyingState);
                 }
 
-                if (superstates.GroupBy(n => n).Any(c => c.Count() > 1))
-                { throw new ArgumentException($"Configuring {State} as a substate of {superstate} creates an illegal nested cyclic configuration."); }
-
+                // The check was OK, we can add this 
                 var superRepresentation = _lookup(superstate);
                 _representation.Superstate = superRepresentation;
                 superRepresentation.AddSubstate(_representation);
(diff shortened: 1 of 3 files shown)


src/Stateless/StateConfiguration.cs
```diff
@@ -508,6 +508,26 @@ public StateConfiguration OnExit(Action<Transition> exitAction, string exitActio
             /// <returns>The receiver.</returns>
             public StateConfiguration SubstateOf(TState superstate)
             {
+                var State = _representation.UnderlyingState;
+
+                // Check for accidental identical cyclic configuration 
+                if (State.Equals(superstate)) { throw new ArgumentException($"Configuring {State} as a substate of {superstate} creates an illegal cyclic configuration."); }
```
nblumhardt: Nit: apart from argument null checks, `if` statements in the repo are mostly over multiple lines. Braces optional, but if they're present they should be on a line each.

src/Stateless/StateConfiguration.cs
```diff
@@ -508,6 +508,26 @@ public StateConfiguration OnExit(Action<Transition> exitAction, string exitActio
...
+
+                // Check for accidental identical nested cyclic configuration 
+                List<TState> superstates = new List<TState>();
```
nblumhardt: Nit: `var superstates = new List<TState>{ State };`

nblumhardt: Thinking through it a little further, could be:

```csharp
var superstates = new HashSet<TState>{ State };
```

The check for membership could then be moved inside the `while` loop as below.

src/Stateless/StateConfiguration.cs
```diff
@@ -508,6 +508,26 @@ public StateConfiguration OnExit(Action<Transition> exitAction, string exitActio
...
+                superstates.Add(State);
+
+                // Build list of super states 
+                var activeRepresentation = _lookup(superstate);
+                while (activeRepresentation.Superstate != null)
+                {
+                    superstates.Add(activeRepresentation.Superstate.UnderlyingState);
```
nblumhardt: Using `HashSet` and moving the check here we can get away from the `GroupBy` below.

```csharp
if (superstates.Contains(activeRepresentation.Superstate.UnderlyingState))
    throw new ArgumentException($"Configuring {State} as a substate of {superstate} creates an illegal nested cyclic configuration.");
```

test/Stateless.Tests/StateMachineFixture.cs
```diff
@@ -359,5 +359,20 @@ public void TheOnTransitionEventFiresBeforeTheOnEntryEvent()
                 Assert.AreEqual(expectedOrdering[i], actualOrdering[i]);
             }
         }
+        [Test]
+        public void CyclicConfigurationDetected()
```
nblumhardt: `DirectCyclicConfigurationDetected()`?

test/Stateless.Tests/StateMachineFixture.cs
```diff
@@ -359,5 +359,20 @@ public void TheOnTransitionEventFiresBeforeTheOnEntryEvent()
...
+        {
+            var sm = new StateMachine<State, Trigger>(State.A);
+
+            Assert.Throws(typeof(ArgumentException),  delegate { sm.Configure(State.A).SubstateOf(State.A); });
```
nblumhardt: Is the old-style `delegate` needed?

```csharp
Assert.Throws(typeof(ArgumentException),  () => { sm.Configure(State.A).SubstateOf(State.A); });
```

test/Stateless.Tests/StateMachineFixture.cs
```diff
@@ -359,5 +359,20 @@ public void TheOnTransitionEventFiresBeforeTheOnEntryEvent()
...
+        }
+        [Test]
```
nblumhardt: Nit: missing newline.

HenningNT:
Thanks for the feedback!
I've fixed the nitpicks, and added a necessary test (which failed initially).
Also switched to Hashset instead of list, it reduced the line count

nblumhardt:
Nice! :-)

HenningNT:
Thanks!

Sendt fra Windows 10-telefonen min

Fra: Nicholas Blumhardt
Sendt: søndag 22. januar 2017 kl. 23.46
Til: dotnet-state-machine/stateless
Kopi: HenningNT; Author
Emne: Re: [dotnet-state-machine/stateless] Adding check for cyclicconfiguration (#125)

Nice! :-)

PR #130, opened 2017-02-19
Support InternalTransition for all TriggerWithParams

### LINKED ISSUE #129: InternalTransition using TriggerWithParameters with more than 1 arg
Hello,
There are trigger three flavors of `TriggerWithParameters` with generic parameters:

- `TriggerWithParameters<TArg0>`
-  `TriggerWithParameters<TArg0, TArg1>`
- `TriggerWithParameters<TArg0, TArg1, TArg2>`

These can be used both when firing events, as well as entering a new state. _However_, the `InternalTransition` only supports the usage of one-argument trigger. My question is: is there any particular reason for this? If not, would it make sense to add support for all `TriggerWithParameters` for `InternalTransition`? If so, I'd be happy to help out.

pardahlman:
This fixes #129 by adding support for internal transition for `TriggerWithParameters` that have up to three generic args.

### DIFF
--- a/src/Stateless/StateConfiguration.Async.cs
+++ b/src/Stateless/StateConfiguration.Async.cs
@@ -73,6 +73,46 @@ namespace Stateless
             }
 
             /// <summary>
+            /// Add an internal transition to the state machine. An internal action does not cause the Exit and Entry actions to be triggered, and does not change the state of the state machine
+            /// </summary>
+            /// <typeparam name="TArg0"></typeparam>
+            /// <typeparam name="TArg1"></typeparam>
+            /// <param name="trigger">The accepted trigger</param>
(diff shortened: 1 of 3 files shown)


HenningNT:
Looks good! Maybe @nblumhardt would like comment?

nblumhardt:
LGTM! 👍

Just one observation - it would be great in the two tests (and in existing test above them too, actually) to assert that the callback/action is actually invoked. Currently they'll assert on the argument values when invoked, but there's no guarantee anything actually happened at all.

pardahlman:
Thanks @nblumhardt see `c476505`. Is that what you're looking for? Also, let me know if you want to squash the commits!

PR #132, opened 2017-02-25
Conditional InternalTransition

Tuan-Minh:
Hi, I have add one InternalTransitionIf for each InternalTransition methods, nothing special about them to write about.

### DIFF
--- a/src/Stateless/InternalTriggerBehaviour.cs
+++ b/src/Stateless/InternalTriggerBehaviour.cs
@@ -10,6 +10,10 @@ namespace Stateless
                 : base(trigger, new TransitionGuard(() => true, "Internal Transition"))
             {
             }
+            public InternalTriggerBehaviour(TTrigger trigger, Func<bool> guard)
+                : base(trigger, new TransitionGuard(guard, "Internal Transition"))
+            {
+            }
             public override bool ResultsInTransitionFrom(TState source, object[] args, out TState destination)
             {
                 destination = source;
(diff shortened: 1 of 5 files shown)


HenningNT:
This looks very good! I haven't tried it yet, but it looks like it can be merged as is.

src/Stateless/InternalTriggerBehaviour.cs
```diff
@@ -10,6 +10,10 @@ public InternalTriggerBehaviour(TTrigger trigger)
                 : base(trigger, new TransitionGuard(() => true, "Internal Transition"))
             {
             }
+            public InternalTriggerBehaviour(TTrigger trigger, Func<bool> guard)
+                : base(trigger, new TransitionGuard(guard, "Internal Transition"))
+            {
+            }
             public override bool ResultsInTransitionFrom(TState source, object[] args, out TState destination)
```
nblumhardt: Missing newlines between the method bodies here.

src/Stateless/InternalTriggerBehaviour.cs
```diff
@@ -10,6 +10,10 @@ public InternalTriggerBehaviour(TTrigger trigger)
                 : base(trigger, new TransitionGuard(() => true, "Internal Transition"))
```
nblumhardt: Chaining constructors would keep things DRYer here:

```
    :this(trigger, () => true)
```

nblumhardt:
Great PR, thanks!

It would cut the amount of duplication in the code quite a bit if we redirected the old methods (without `guard`s) to call the new methods. I made a comment regarding the constructor of `InternalTriggerBehavior`, but I think the same applies to `InternalPermitIf()` and so-on.

Tuan-Minh:
@nblumhardt I have address your comment regarding duplicitous code. Silly me!

nblumhardt:
Great! Thanks.

PR #137, opened 2017-02-28
Replace magic string usages with nameof()

Tuan-Minh:
Nothing significant, just all nameof() for all Enforce.ArgumentNotNull calls.


nblumhardt:
Great, thank you!

PR #141, opened 2017-03-24
Update to RTM dotnet tooling/VS2017

nblumhardt:
Now requires VS 2017, but the availability of (free) VS Community and the .NET CLI should keep contribution open to anyone.

This PR removes the "old-style" PLIB target, which should be covered on all supported platforms by netstandard-1.0.

(Also updates TFM to 4.0.0 - see #103).


nblumhardt:
(Merging so that forthcoming PRs aren't impossible to merge, given the churn in this giant-but-boring changeset.)

PR #145, opened 2017-03-24
More changes from dev

HenningNT:
More changes from dev


HenningNT:
That was the wrong way round...

Issue #143, opened 2017-03-24
Reflection API remaining tasks


nblumhardt:
From #103 

 - [x] `Superstate` needs to be lazily linked up as per the transitions, so that `x.Superstate` is a fully-populated info object
 - [x] Use a common base class for all transitions, so that e.g. `DynamicTransitionInfo` and `TransitionInfo` can share  common attributes
 - [x] `DynamicTransitionInfo.Destination` -> `DestinationDescription`
 - [x] Consider `ActionInfo` to represent actions; in the process we may be able to standardize how `Description` values are handled (can we unify actions and guards as `InvocationInfo` or similar?)
 - [x] Update the DotGraph HTML work and merge
 - [x] `TriggerInfo.Value` -> `UnderlyingTrigger`
 - [x] `StateType` and `TriggerType` should be on `StateMachineInfo`
 - [x] `TransitionInfo.GuardDescription` should be plural - multiple guards can be attached to a single trigger

HenningNT:
I don't like to bring bad news, especially considering all the amazing work that has been done... but I had a look this morning, and I couldn't get it to work. I tried it on the Bug example to get a DotGraph, and this resulted in a StackOverflowException.

CreateStateInfo has a recursive call when it generates a list of substates. A superstate always has substates, and so it then calls CreateStateInfo on each substate. The substates have superstates, so it calls CreateStateInfo on the superstate. This goes on until the stack is full.

I did a quick hack, which did produce graphs for the Bug example and the PhoneCall example:

        internal static StateInfo CreateStateInfo<TState, TTrigger>(StateMachine<TState, TTrigger>.StateRepresentation stateReperesentation)
        {
            if (stateReperesentation == null)
                throw new ArgumentException(nameof(stateReperesentation));

            var substates = stateReperesentation.GetSubstates().Select(s => CreateSubStateInfo(s)).ToList();
            var entryActions = stateReperesentation.EntryActions.Select(e => e.ActionDescription).ToList();
            var exitActions = stateReperesentation.ExitActions.Select(e => e.ActionDescription).ToList();
            var ignoredTriggers = FindIgnoredTriggers(stateReperesentation);

            StateInfo superstate = CreateSuperStateInfo(stateReperesentation.Superstate, substates);

            return new StateInfo(stateReperesentation.UnderlyingState, typeof(TState), superstate, substates, entryActions, exitActions, ignoredTriggers);
        }
[...]

PR #147, opened 2017-03-27
Extracting common abstract base class for transition info

HenningNT:
Extracting common abstract base class for transition info.
Added new abstract base class for common properties.
Updating tests and other code as needed.


HenningNT:
... as mentioned in #143

src/Stateless/TransitionBase.cs
```diff
@@ -0,0 +1,28 @@
+﻿namespace Stateless
```
nblumhardt: Looks like this file might be a leftover?

src/Stateless/Reflection/TransitionInfo.cs
```diff
@@ -1,52 +1,18 @@
...
-        /// </summary>
-        public StateInfo DestinationState { get; private set; }
-
+        public TriggerInfo Trigger { get; internal set; }
 
         /// <summary>
         /// Description of provided guard clause, if any.
         /// </summary>
-        public string GuardDescription { get; private set; }
+        public string GuardDescription { get; internal set; }
```
nblumhardt: Could this be a `protected set`?

src/Stateless/TransitionBase.cs
HenningNT: Yes, that should be there... It doesn't show up in Solution Explorer, I'll remove this!

nblumhardt:
Looks good!

I'm a little uneasy with the name `ExternalTransitions` - but I realise you're sticking with the standard terminology and I can't come up with a better alternative. It's really odd how (at least in [the UML lexicon](https://en.wikipedia.org/wiki/UML_state_machine#Local_versus_external_transitions)), _external_ transitions are the opposite of _local transitions_, rather than _internal transitions_ (which also appear in the lexicon). Eek! :-)

Any other possible options for the name? If not, this all seems ready to go.

src/Stateless/Reflection/TransitionInfo.cs
HenningNT: Yes, it can :-)

HenningNT:
Naming things are hard...
I just had a look at the latest UML specification:

> 14.2.3.8.1 Transition kinds relative to source
> The semantics of a Transition depend on its relationship to its source Vertex. Three different possibilities are defined,
> depending on the value of the Transition’s kind attribute:
> • kind = external means that the Transition exits its source Vertex. If the Vertex is a State, then executing this
> Transition will result in the execution of any associated exit Behavior of that State.
> • kind = local is the opposite of external, meaning that the Transition does not exit its containing State (and,
> hence, the exit Behavior of the containing State will not be executed). However, for local Transitions the target
> Vertex must be different from its source Vertex. A local Transition can only exist within a composite State.
> • kind = internal is a special case of a local Transition that is a self-transition (i.e., with the same source and
> target States), such that the State is never exited (and, thus, not re-entered), which means that no exit or entry
> Behaviors are executed when this Transition is executed. This kind of Transition can only be defined if the
> source Vertex is a State.

I think the names are fairly OK, and also realize that we should maybe rename InternalTransition() to something else, maybe PermitInternal() and add PermitLocal() ;-)

nblumhardt:
Thanks for the follow-up. There is some appeal in renaming `InternalTransition` to `PermitInternal()`, or something along those lines.

I think `ExternalTransition` is actually a misnomer now, since because of the implicit hierarchy, this will also include local transitions. (I guess this is what you're getting at with `PermitLocal()`).

I have to run, but I'll give it some more thought!

HenningNT:
Something along those lines, yes. If we want to conform to the UML naming, then we could either use PermitExternal, PermitLocal and PermitInternal. Or just go with Permit, and let the Permit method figure it out. The latter has a simpler API, but having explicit Permits makes it quite clear to the users what they are asking for.

For the reflection stuff it might be worth splitting up transitions into ExternalTransition, LocalTransition and InternalTransition?

nblumhardt:
I don't think the switch to `PermitExternal`/`PermitLocal` would work so well with the current API, since the difference between the two is implied by the state hierarchy. Maybe there's too much water under the bridge to be fully aligned with the UML terms.

From the model's point of view, would it make sense to say we have two kinds of transitions - those we know the destination state of (...fixed...?) and those we don't (dynamic, today). I think the reflection end of things we could have `TransitionInfo` (abstract), `FixedTransitionInfo` and `DynamicTransitionInfo`?

Then at the API, we just try to fit `InternalTransition` in among the other `Permit` variants:

```csharp
    .Permit(Trigger.Closed, State.Closed)              // FixedTransitionInfo
    .PermitDynamic(Trigger.Closed, t => State.Closed)  // DynamicTransitionInfo
    .PermitReentry(Trigger.Assigned)                   // FixedTransitionInfo
    .PermitInternal(Trigger.Assigned, t => Act())      // FixedTransitionInfo
```

(I think this is a breaking change we could afford to take in the new major version.)

The alternative would be to model internal transitions as "actions", so `PermitAction()` or just `OnTrigger()`. I'm okay with the first, would need to think more carefully about the second.

What do you think?

HenningNT:
I like the Dynamic/Fixed TransitionInfo! I'll try that, and see what happens :-)

We can discuss changing the API later 👍

HenningNT:
ok, so I tried to simplify it a bit... maybe I did too many changes?

test/Stateless.Tests/DotGraphFixture.cs
```diff
@@ -33,7 +33,7 @@ public void SimpleTransition_DotGraph()
 
             sm.Configure(State.A)
                 .Permit(Trigger.X, State.B);
-
+            var resukt = DotGraphFormatter.Format(sm.GetInfo());
```
nblumhardt: This doesn't look like it belongs here :-)

nblumhardt:
Great, thanks for the update! Still trying to get some time spare to run back through it :-)

test/Stateless.Tests/DotGraphFixture.cs
HenningNT: You're right!

src/Stateless/Reflection/StateInfo.cs
```diff
@@ -130,12 +122,7 @@ public class StateInfo
         /// <summary>
         /// Transitions defined for this state.
         /// </summary>
-        public IEnumerable<TransitionInfo> Transitions { get; private set; }
-
-        /// <summary>
-        /// Transitions defined for this state internally.
-        /// </summary>
-        public IEnumerable<TransitionInfo> InternalTransitions { get; private set; }
+        public IEnumerable<FixedTransitionInfo> ExternalTransitions { get; private set; }
```
nblumhardt: Should/could this be `FixedTransitions`?

src/Stateless/Reflection/StateInfo.cs
```diff
@@ -130,12 +122,7 @@ public class StateInfo
         /// <summary>
         /// Transitions defined for this state.
         /// </summary>
-        public IEnumerable<TransitionInfo> Transitions { get; private set; }
```
nblumhardt: Could this remain, but simply return `FixedTransitions.Concat(DynamicTransitions)`? In .NET reflection, we have this pattern in a few places e.g. https://msdn.microsoft.com/en-us/library/system.reflection.typeinfo.declaredmembers(v=vs.110).aspx)

src/Stateless/Reflection/StateInfo.cs
HenningNT: Yup!

src/Stateless/Reflection/StateInfo.cs
HenningNT: Sure we can, some might find it useful.

nblumhardt:
Cool :-) ... good to go? 🚢 🚢 🚢

HenningNT:
I think so! 👍

nblumhardt:
Great! :-)

Circling back... What do you think about a possible rename from `InternalTransition()` to `PermitSomething()`?

HenningNT:
Renaming `InternalTransition `to `PermitInternal `seems like a sensible move.

PR #149, opened 2017-04-04
Moved StateType and TriggerType into StateMachineInfo, removed TriggerInfo

betty-crokker:


betty-crokker:
This is in response to items listed in #143

nblumhardt:
Thanks Chris!

The shift onto `StateMachineInfo` of `StateType` and `TriggerType` looks like a winner 👍 

I can definitely see us considering the `TriggerInfo` removal, but I'm not 100% sure yet that it's the way to go - although the type doesn't carry any data other than the `UnderlyingTrigger` right now, it does make the API a little more self-documenting than if we were to use `object` throughout.

Can we get the one change through, and leave the `TriggerInfo` change to bake a little longer?

Cheers!
Nick

betty-crokker:
@nblumhardt I can sort of see what you mean, but I'm not really convinced that a property named Trigger.Value that returns an object is more self-documenting than a property named Trigger that returns an object, especially since we don't talk about a trigger "value" anywhere else in the code.  My two cents worth: bite the bullet, generic-ize TriggerInfo to TriggerInfo<TTrigger> and then the Trigger property is of type TTrigger instead of object.  Now that's self-documenting!

nblumhardt:
Thanks for the follow-up! (Aside: the property would be `UnderlyingValue` - there's another checklist item for renaming it).

I'm not suggesting that we never make this change - only that we split it from the `StateType`/`TriggerType` discussion and let things bake a little first :-)

betty-crokker:
@nblumhardt OK, there were only a couple tweaks from the main dev needed, I'll submit the "remove TriggerInfo class" as a separate PR.

betty-crokker:
Ok, that's confusing, still getting used to github.  That commit "Remove unnecessary TriggerInfo" was not supposed to be part of this PR.

nblumhardt:
Great, thanks Chris!

PR #154, opened 2017-04-06
Create new ActionInfo class, start insinuating it

betty-crokker:
There's a lot here, let me take you through my thought process:

1) We want an ActionInfo class that formatters like dotgraph can use to get the names of EntryActions, ExitActions, etc.
1a) Right now, if the user creates an entry action that is a delegate or lambda, the resulting chart uses the compiler generated function name (e.g. "<.ctor>b__2_0") which is not terribly human-friendly.

2) In order for the ActionInfo class to handle the delegate/lambda name issue, we need to be able to distinguish action descriptions the user created vs. action descriptions that were automatically generated from the method name
2a) For example: in our ActionInfo class, if we reject any action descriptions containing angle brackets (assuming that indicates the function was a delegate or lambda), but it's actually a user-specified description like "Enable button if count > 3", the user will be unhappy.

3) Right now, the code that detects that the user has not specified an action description, and that we should generate an action description based on the function name, is in many locations.
3a) So instead, have the functions in StateConfiguration just pass entryActionDescription on, without worrying about null values.
3b) Deal with (entryActionDescription == null) in one place: the EntryActionBehaviour constructor.

4) The ActionInfo class should work for any type of action behaviour, not just EntryActionBehaviour.
4a) Create a base class named ActionBehaviour, have that be what ActionInfo responds to.
4b) For backward compatibility, make it so the ActionDescription property in this new ActionBehaviour class works the same as it has in the past.
4c) Add some new knowledge (a property named "ActionDescriptionIsCalculated") that ActionInfo can use to determine whether angle brackets are bad (compiler-generated function names) or good (user-specified action descriptions)

As I said, this is a lot, so right now I've only implemented this for EntryActionBehaviour.  If it's agreed this is the right direction to go, I'll continue by having the other *ActionBehaviour classes derive from ActionBehaviour.


nblumhardt:
Great, thanks Chris.

It's a bit hard to review because the branch you're sending the PR from also contains the TriggerType/StateType changes and history (ideally you want to branch each PR off the dev branch individually).

To get back to convergent history, after taking a backup copy of files from your workspace, I think you need to:

1. add upstream as a remote, if you haven't already:

```
git remote add upstream https://github.com/dotnet-state-machine/stateless
```

2. pull in and rebase on top of the upstream dev branch

```
git pull upstream dev --rebase
```

3. fix any merge conflicts that show up (I often use git extensions for this)

4. push back to your fork:

```
git push origin Actions --force
```

Cheers!
Nick

PR #157, opened 2017-04-06
Remove the no-longer-needed TriggerInfo class

betty-crokker:
TriggerInfo, after moving the TriggerType property up to StateMachineInfo, was left with only a single member, an object named Value. This PR removes the TriggerInfo class completely, and so the TransitionInfo class now has an "object UnderlyingTrigger" instead of a "TriggerInfo Trigger".

Readability and self-documentability are not decreased; code, instead of referring to "transitionInfo.Trigger.Value", now refers to "transitionInfo.UnderlyingTrigger".


betty-crokker:
@nblumhardt what do you think?

betty-crokker:
Here's an example of why I want to remove TriggerInfo - here's a snippet from one of the unit tests :

            foreach (FixedTransitionInfo transition in binding.FixedTransitions)
            {
                Assert.True(transition .Trigger.UnderlyingTrigger is Trigger);
                Assert.Equal(Trigger.X, (Trigger)transition .Trigger.UnderlyingTrigger);

When you read this code, it looks like there's one trigger ("Trigger"), and underneath it is some other trigger ("UnderlyingTrigger").  If we renamed "UnderlyingTrigger" to "Value", it's a little better:

            foreach (FixedTransitionInfo transition in binding.FixedTransitions)
            {
                Assert.True(transition .Trigger.Value is Trigger);
                Assert.Equal(Trigger.X, (Trigger)transition .Trigger.Value);

But it's still a little confusing that a trigger has a value which is somehow different from the trigger itself.  The most readable is without the TriggerInfo class:

            foreach (FixedTransitionInfo trans in binding.FixedTransitions)
            {
                Assert.True(transition .Trigger is Trigger);
                Assert.Equal(Trigger.X, (Trigger)transition .Trigger);

My two cents worth :-) !

betty-crokker:
@nblumhardt Any new thoughts on this?  I really think we should remove it, but am cool with leaving it in if you feel equally strongly the other direction.  Just want to make a decision and move on :-) !

nblumhardt:
Thanks for the nudge. I still feel like this needs time to bake - it won't be lost here, and we should have some more experience as consumers of the API soon to feed into this.

betty-crokker:
@nblumhardt @HenningNT Well baked now ;-)  

There's still no reason to have this class, but if you're still uncomfortable removing it let's close this PR, it's not that big of a deal to me.

nblumhardt:
Thanks for the nudge @betty-crokker. I'm still leaning towards the explicit type, so let's close this and see how it goes. Cheers!

PR #158, opened 2017-04-06
Create new ActionInfo class, start insinuating it

betty-crokker:
There's a lot here, let me take you through my thought process:

1) We want an ActionInfo class that formatters like dotgraph can use to get the names of EntryActions, ExitActions, etc.
1a) Right now, if the user creates an entry action that is a delegate or lambda, the resulting chart uses the compiler generated function name (e.g. "<.ctor>b__2_0") which is not terribly human-friendly.

2) In order for the ActionInfo class to handle the delegate/lambda name issue, we need to be able to distinguish action descriptions the user created vs. action descriptions that were automatically generated from the method name
2a) For example: in our ActionInfo class, if we reject any action descriptions containing angle brackets (assuming that indicates the function was a delegate or lambda), but it's actually a user-specified description like "Enable button if count > 3", the user will be unhappy.

3) Right now, the code that detects that the user has not specified an action description, and that we should generate an action description based on the function name, is in many locations.
3a) So instead, have the functions in StateConfiguration just pass entryActionDescription on, without worrying about null values.
3b) Deal with (entryActionDescription == null) in one place: the EntryActionBehaviour constructor.

4) The ActionInfo class should work for any type of action behaviour, not just EntryActionBehaviour.
4a) Create a base class named ActionBehaviour, have that be what ActionInfo responds to.
4b) For backward compatibility, make it so the ActionDescription property in this new ActionBehaviour class works the same as it has in the past.
4c) Add some new knowledge (a property named "ActionDescriptionIsCalculated") that ActionInfo can use to determine whether angle brackets are bad (compiler-generated function names) or good (user-specified action descriptions)

5) Implement ideas from @nblumhardt  about using the same class for Guard Conditions too (so rename ActionBehaviour to InvocationInfo and bring GuardCondition.cs into the family)

As I said, this is a lot, so right now I've only implemented this for EntryActionBehaviour. If it's agreed this is the right direction to go, I'll continue by having the other *ActionBehaviour classes derive from InvocationInfo.


betty-crokker:
Has anyone had a chance to look at this idea?  If it's agreed upon, I'll finish implementing it.

betty-crokker:
@HenningNT you had suggested using "the text "(lambda)" or something similar as names for the actions", I used "Function" but put it in a static member function so it was easy to change.  Maybe we want to make it a static property so it's easier for individual developers to make their own choice?  

I wasn't crazy about "lambda" because in my world, the state diagram may be viewed by non-developers who have no idea what a "lambda" is, but then as I think about it more, just saying "Function" doesn't really help much either, it's awfully non-specific.

Maybe the exact word choice isn't important, if users are really worried about the text displayed in the graph, they will take the time to specify the 'actionDescription' parameter?

Whaddayathink?

src/Stateless/EntryActionBehaviour.cs
```diff
@@ -31,6 +30,9 @@ public Sync(Action<Transition, object[]> action, string actionDescription) : bas
                     _action = action;
                 }
 
+                internal override string MethodName {  get { return _action.TryGetMethodName(); } }
```
nblumhardt: Since the class itself is internal (via the parent) these could just be `public`.

src/Stateless/StateConfiguration.cs
```diff
@@ -415,7 +415,7 @@ public StateConfiguration OnEntry(Action entryAction, string entryActionDescript
                 Enforce.ArgumentNotNull(entryAction, nameof(entryAction));
                 return OnEntry(
                     t => entryAction(),
-                    entryActionDescription ?? entryAction.TryGetMethodName());
+                    entryActionDescription);
```
nblumhardt: Because the action is wrapped by the lambda expression `t => ...`, we'll lose the original method name with this change.

nblumhardt: Could the ultimate private implementation of `OnEntry` accept an `ActionInfo` like this?

```csharp
return OnEntry(t => entryAction(), ActionInfo.Create(entryAction, entryActionDescription));
```

betty-crokker:
@nblumhardt I agree with what you're saying about not calling it InvocationInfo, but I'm not crazy about renaming it to Invocation since the class doesn't actually store the invocation.  Having said that, I'm not coming up with anything I like much better ... what do you think about MethodDescription for the class name?

src/Stateless/EntryActionBehaviour.cs
betty-crokker: I made those methods "internal" in the abstract base class (formerly named InvocationInfo) for when it's inherited by GuardCondition (where there aren't these internal classes).  I think that in order to make MethodName public here, I would have to make it public in InvocationInfo, and then it would end up public in GuardCondition ...

src/Stateless/StateConfiguration.cs
betty-crokker: That makes a lot of sense!  I've started using that idea, you can see in my latest commit.

src/Stateless/MethodDescription.cs
```diff
@@ -0,0 +1,51 @@
...
+    internal class MethodDescription
+    {
+        readonly string _description;                     // _description can be null if user didn't specify a description
+
+        public enum Timing
+        {
+            Synchronous,
+            Asynchronous
+        }
+        Timing _timing;
```
nblumhardt: `readonly`?

src/Stateless/MethodDescription.cs
```diff
@@ -0,0 +1,51 @@
+﻿using System;
+
+namespace Stateless
+{
+    internal class MethodDescription
+    {
+        readonly string _description;                     // _description can be null if user didn't specify a description
+
+        public enum Timing
```
nblumhardt: Nice touch!

src/Stateless/MethodDescription.cs
```diff
@@ -0,0 +1,51 @@
...
+            Asynchronous
+        }
+        Timing _timing;
+
+        public static MethodDescription Create(Delegate method, string description, Timing timing = Timing.Synchronous)
+        {
+            return new MethodDescription(method?.TryGetMethodName(), description, timing);
+        }
+
+        private MethodDescription(string methodName, string description, Timing timing)      // description can be null if user didn't specify a description
```
nblumhardt: No `private` modifier needed here.

src/Stateless/MethodDescription.cs
```diff
@@ -0,0 +1,51 @@
...
+            MethodName = methodName;
+            _description = description;
+            _timing = timing;
+        }
+
+        /// <summary>
+        /// The name of the invoked method.  If the method is a lambda or delegate, the name will be a compiler-generated
+        /// name that is often not human-friendly (e.g. "(.ctor)b__2_0" except with angle brackets instead of parentheses)
+        /// </summary>
+        internal string MethodName { get; private set; }
```
nblumhardt: These `internal` methods can all be `public`, since it's an `internal` class.

src/Stateless/MethodDescription.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// <summary>
+        /// Returns true if the value returned from ActionDescription is calculated from the action's method,
+        /// false if it is a value specified by the user.
+        /// </summary>
+        internal bool DescriptionIsCalculated { get { return (_description == null);  } }
+
+        /// <summary>
+        /// Returns true if the method is invoked asynchronously.
+        /// </summary>
+        internal bool IsAsync => (_timing == Timing.Asynchronous);
```
nblumhardt: Looking now at how this is used, would it be simpler to just specify `isAsync` in the constructor, and avoid the `Timing` enum?

src/Stateless/Reflection/MethodInfo.cs
```diff
@@ -0,0 +1,54 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Stateless.Reflection
+{
+    /// <summary>
+    /// Describes an internal method through the reflection API.
+    /// </summary>
+    public class MethodInfo
```
nblumhardt: Unfortunately the name `MethodInfo` overlaps with `System.Reflection.MethodInfo`; the namespace does disambiguate them, but seeing `MethodInfo` in any reflection-related code would immediately suggest the framework class to me. What other contenders do we have for this name? `InvocationInfo` was one, any others?

src/Stateless/Reflection/MethodInfo.cs
```diff
@@ -0,0 +1,54 @@
...
+            string descrip;
+
+            descrip = action.Description;
+
+            // If the action description was calculated based on the method name, and the method was a delegate or lambda,
+            // don't use the compiler-generated method name.
+            if ( (action.DescriptionIsCalculated) && (descrip.IndexOfAny(new char[] { '<', '>' }) >= 0) )
+                descrip = _defaultFunctionName;
+
+            MethodInfo actionInfo = new MethodInfo(descrip, action.IsAsync);
```
nblumhardt: Nit: `var` works here.

src/Stateless/Reflection/MethodInfo.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// </summary>
+    public class MethodInfo
+    {
+        static readonly string _defaultFunctionName = "Function";
+
+        internal static MethodInfo CreateMethodInfo(MethodDescription action)
+        {
+            string descrip;
+
+            descrip = action.Description;
```
nblumhardt: `var descrip = action.Description`?

src/Stateless/StateConfiguration.Async.cs
```diff
@@ -236,7 +236,7 @@ public StateConfiguration OnEntryAsync(Func<Task> entryAction, string entryActio
                 Enforce.ArgumentNotNull(entryAction, nameof(entryAction));
                 return OnEntryAsync(
                     t => entryAction(),
-                    entryActionDescription ?? entryAction.TryGetMethodName());
+                    entryActionDescription);
```
nblumhardt: We're still losing the method name in these calls (I think the change applied to the sync class is needed here, too).

src/Stateless/EntryActionBehaviour.cs
```diff
@@ -26,11 +27,14 @@ public class Sync : EntryActionBehavior
...
                 readonly Action<Transition, object[]> _action;
 
-                public Sync(Action<Transition, object[]> action, string actionDescription) : base(actionDescription)
+                public Sync(Action<Transition, object[]> action, MethodDescription description) : base(description)
                 {
                     _action = action;
                 }
 
+                //internal override string MethodName {  get { return _action.TryGetMethodName(); } }
+                //internal override bool IsAsync {  get { return false;  } }
```
nblumhardt: Nit: are these lines needed/can they be removed?

src/Stateless/EntryActionBehaviour.cs
```diff
@@ -10,14 +10,15 @@ public partial class StateMachine<TState, TTrigger>
...
 
-            protected EntryActionBehavior(string actionDescription)
+            protected EntryActionBehavior(MethodDescription description)
             {
-                _actionDescription = Enforce.ArgumentNotNull(actionDescription, nameof(actionDescription));
+                _description = description;
             }
 
-            internal string ActionDescription { get { return _actionDescription; } }
+            // Rename base class 'Description' for backward compatibility
```
nblumhardt: Is this comment accurate? The current class doesn't have a base class. Since it's not a public type, do we need to worry about compatibility here at all?

src/Stateless/MethodDescription.cs
betty-crokker: I went back and forth on that ... I'm not a big fan of bools as parameters - when you look at the calling code, you can't tell what the "true" or "false" means.  Having an enum there makes it really clear.

src/Stateless/Reflection/MethodInfo.cs
betty-crokker: I simplified this even further, there's no reason to have that intermediate value at all, just return the new'ed MethodInfo.

src/Stateless/EntryActionBehaviour.cs
betty-crokker: Gone!

src/Stateless/EntryActionBehaviour.cs
betty-crokker: Good point.  Comment updated.

src/Stateless/StateConfiguration.Async.cs
betty-crokker: Good catch!  I think I've got them all now.

src/Stateless/MethodDescription.cs
betty-crokker: Done!

src/Stateless/MethodDescription.cs
betty-crokker: OK

src/Stateless/MethodDescription.cs
betty-crokker: Yes they can ... but I'm not sure it increases the readability of the code.  It's an interesting problem, you might consider it a limitation of the language.  If you read a method that says it's public, you don't know what that means until you look up at the class definition too.  I guess I went with defining the property as internal so it was more obvious what its true accessibility is.

src/Stateless/Reflection/MethodInfo.cs
betty-crokker: Argh.  Yeah, that's not good to overlap with the system class.  InvocationInfo is doable, but thinking outside the box for a second - can we get rid of the MethodInfo class completely and just add the "calculate the description" method to MethodDescription, and use it for both the main state machine and the reflection API?

betty-crokker:
I've started putting in some unit tests, and have to think about that.  I'd like to test every function that generates one of these MethodInfo objects (whatever we end up calling them), and for each such function I want to test 4 situations - (1) named method, (2) named method with description, (3) lambda, and (4) lambda with description.  That's going to be a lot of tests!

nblumhardt:
Awesome. I will need to find some time later in the week to run though again, apologies in advance if things are a bit sluggish (lots going on :-))

betty-crokker:
@nblumhardt Poke!  :bowtie:

example/ReflectionExample/Program.cs
```diff
@@ -0,0 +1,73 @@
+﻿using System;
```
nblumhardt: Since BugTrackerExample already exports to DOT graph, could we just tweak that one rather than add a new one to demonstrate reflection capabilities? https://github.com/dotnet-state-machine/stateless/blob/dev/example/BugTrackerExample/Program.cs#L23

src/Stateless/ActivateActionBehaviour.cs
```diff
@@ -8,15 +8,16 @@ public partial class StateMachine<TState, TTrigger>
...
 
-            protected ActivateActionBehaviour(TState state, string actionDescription)
+            protected ActivateActionBehaviour(TState state, MethodDescription actionDescription)
             {
                 _state = state;
                 _actionDescription = Enforce.ArgumentNotNull(actionDescription, nameof(actionDescription));
             }
 
-            internal string ActionDescription { get { return _actionDescription; } }
+            internal string ActionDescription { get { return _actionDescription.Description; } }
```
nblumhardt: Do we still need this helper property?

src/Stateless/DeactivateActionBehaviour.cs
```diff
@@ -11,15 +11,16 @@ public partial class StateMachine<TState, TTrigger>
...
 
-            protected DeactivateActionBehaviour(TState state, string actionDescription)
+            protected DeactivateActionBehaviour(TState state, MethodDescription actionDescription)
             {
                 _state = state;
                 _actionDescription = Enforce.ArgumentNotNull(actionDescription, nameof(actionDescription));
             }
 
-            internal string ActionDescription { get { return _actionDescription; } }
+            internal string ActionDescription { get { return _actionDescription.Description; } }
```
nblumhardt: Also probably unneeded here.

src/Stateless/ActivateActionBehaviour.cs
```diff
@@ -8,15 +8,16 @@ public partial class StateMachine<TState, TTrigger>
...
+            internal MethodDescription MethodDescription => _actionDescription;
```
nblumhardt: This can be `public`, since it's on an `internal` type.

src/Stateless/EntryActionBehaviour.cs
```diff
@@ -10,14 +10,18 @@ public partial class StateMachine<TState, TTrigger>
...
+            protected EntryActionBehavior(MethodDescription description)
             {
-                _actionDescription = Enforce.ArgumentNotNull(actionDescription, nameof(actionDescription));
+                _description = description;
             }
 
-            internal string ActionDescription { get { return _actionDescription; } }
+            // Returns a description of the entry action.  If the caller specified a description, it
+            // is returned, else the name of the method itself.
+            public string ActionDescription => _description.Description;
```
nblumhardt: Also probably no longer required.

src/Stateless/ActivateActionBehaviour.cs
nblumhardt: Also, can we perhaps just call these `Description`?

src/Stateless/MethodDescription.cs
nblumhardt: Thanks for the explanation. I don't think the rest of the code uses this convention, except perhaps in one or two places where we might be inconsistent. I'd rather we keep it consistently `public` here.

src/Stateless/Reflection/MethodInfo.cs
nblumhardt: Sounds good 👍  though it'd be nice for it to appear only under the _Reflection_ part of the public API (so we'd still need to rename it `InvocationInfo` or similar).

src/Stateless/Reflection/StateInfo.cs
```diff
@@ -102,14 +103,24 @@ public class StateInfo
...
         /// <summary>
-        /// Actions that are defined to be exectuted on state-entry.
+        /// Actions that are defined to be executed on state-entry.
+        /// </summary>
+        public IEnumerable<MethodInfo> EntryActions { get; private set; }
+
+        /// <summary>
+        /// Actions that are defined to be executed on activation.
+        /// </summary>
+        public IEnumerable<MethodInfo> ActivateActions { get; private set; }
```
nblumhardt: Great catch 👍

src/Stateless/TriggerBehaviour.cs
```diff
@@ -9,38 +9,42 @@ public partial class StateMachine<TState, TTrigger>
     {
         internal abstract class TriggerBehaviour
         {
-            readonly TransitionGuard _guard;
+            readonly TransitionGuard _guard;        // null if there is no guard function
```
nblumhardt: Perhaps, since `TransitionGuard.Conditions` now can be empty, we could default this to a `TransitionGuard` with no conditions, and then delegate some of the functionality below `Guards`, `GuardConditionsMet`, `UnmetGuardConditions` to it so that we don't need null checks in this class?

example/ReflectionExample/Program.cs
betty-crokker: On my machine, BugTrackerExample creates a DLL, not an EXE, so I can't actually run it from the command line.  Something with the target framework being .NETCoreApp?  In any case, I don't need this so I've removed it.

src/Stateless/ActivateActionBehaviour.cs
betty-crokker: I don't ;-) .  I left it in there for backwards compatibility.  Should I remove it?

src/Stateless/ActivateActionBehaviour.cs
betty-crokker: ActionDescription removed; MethodDescription renamed to Description

src/Stateless/DeactivateActionBehaviour.cs
betty-crokker: Agreed

src/Stateless/EntryActionBehaviour.cs
betty-crokker: Agreed

src/Stateless/Reflection/StateInfo.cs
betty-crokker: @nblumhardt I've updated the access from internal to public for consistency, but don't understand your comment about MethodInfo.   MethodInfo is only used in the Reflection part of the public API.  Are you looking at MethodDescription?

src/Stateless/TriggerBehaviour.cs
betty-crokker: Done!

betty-crokker:
@nblumhardt What do you think?

example/ReflectionExample/Program.cs
nblumhardt: Files are still left lingering here - needs to be removed via the filesystem.

src/Stateless/Reflection/MethodInfo.cs
nblumhardt: Still work-in-progress on this one?

betty-crokker:
@nblumhardt I merged MethodInfo into MethodDescription and got rid of MethodInfo, and removed the files for ReflectionExample.

src/Stateless/MethodDescription.cs
```diff
@@ -0,0 +1,71 @@
+﻿using System;
+
+namespace Stateless
+{
+    /// <summary>
+    /// Describes a method - either an action (activate, deactivate, etc.) or a transition guard
+    /// </summary>
+    public class MethodDescription
```
nblumhardt: Since this is only exposed in the reflection API, can we please move it to the _Stateless.Reflection_ namespace and go with the `*Info` naming pattern? I know it's used under the hood outside of that context, but for a consumer this is going to be part of the reflection subsystem only.

`InvocationInfo` was the best non-conflicting name I could come up with - perhaps we can go with that until a better option emerges?

src/Stateless/MethodDescription.cs
```diff
@@ -0,0 +1,71 @@
...
+    {
+        readonly string _description;                     // _description can be null if user didn't specify a description
+
+        /// <summary>
+        /// Is the method synchronous or asynchronous?
+        /// </summary>
+        public enum Timing
```
nblumhardt: Since the class is now public, this should be `internal` or else it will form part of the public API.

src/Stateless/Reflection/StateInfo.cs
```diff
@@ -27,7 +25,14 @@ public class StateInfo
                 }
             }
 
-            return new StateInfo(stateReperesentation.UnderlyingState, entryActions, exitActions, ignoredTriggers);
+            StateInfo stateInfo = new StateInfo(stateReperesentation.UnderlyingState, ignoredTriggers);
+
+            stateInfo.EntryActions = stateReperesentation.EntryActions.Select(e => e.Description).ToList();
```
nblumhardt: Can we stick with the initialized-by-constructor pattern here? It does result in quite a large constructor argument list, but makes it less likely we'll forget to pass these through, and thus end up with null-valued properties, somewhere else.

src/Stateless/MethodDescription.cs
betty-crokker: MethodDescription is what the Stateless classes (ActivateActionBehaviour, EntryActionBehaviour, etc.) use internally to keep track of information about the actions; even though it's only exposed in the reflection API it's used all over the place in the base namespace so I don't think it makes sense to move it to Reflection.

src/Stateless/MethodDescription.cs
betty-crokker: Done.

betty-crokker:
@nblumhardt I made 2 of your 3 changes, the other one I commented on ...

nblumhardt:
Thanks Chris!

The external API organisation usually takes precedence over the internal uses of a type. It's not really consequential to a user whether Stateless uses `MethodDescription` internally in any particular place - building a coherent model for them around their use of the type (as part of the reflection API) is the foremost goal.

It's also possible that the internal organisation will change in the future, while it's much harder to make changes to the public API surface.

Does this explain the thinking? Cheers!

betty-crokker:
I'm mostly following you but not completely ... so you are proposing to move MethodDescription into the Reflection namespace (possibly renaming it too), but still have ActivateActionBehaviour, EntryActionBehaviour, etc. have member variables of that type for storing the information about the methods?

nblumhardt:
Hi Chris - yes, that's correct :+1:

betty-crokker:
Done!  I went with calling it "Reflection.InvocationInfo" everywhere rather than doing a "using Stateless.Reflection", what do you think?

nblumhardt:
Awesome 👍 

Great, PR, thanks!

(We churned on this one a bit, if you decide to take a shot at another, shall we try to map things out together in an issue/discussion more closely next time to try to reduce the round trips? Cheers!)

Issue #138, opened 2017-03-08
Getting reasons for UnmetGuardConditions


KrauseT:
In my scenario, I would like to know the transitions from the statemachine that can not be executed in the current state. In addition, I would also like to get the description of the failed conditions (as a list) to show the user why he can not perform this transition. 
I can imagine that other users have similar scenarios.
Issue #55 has already covered something similar, but an exception with the description is only returned when the transition is executed.

Here is a small sample configuration:
```C#
sm.Configure(State.A)
  .PermitIf(Trigger.Request, State.B,
    new Tuple<Func<bool>, string>(() => false, "calculated vaule should be in range ...."),
    new Tuple<Func<bool>, string>(() => true, "do it"))
  .PermitIf(Trigger.Abort, State.C,
    new Tuple<Func<bool>, string>(() => true, "make it"))
  ;
```

Now the user can return all executable actions. But he has no clue why he can not perform the "_Request_" action (and no button on the surface is visible or this button is disabled).

With a few small adjustments, it is easy to provide this information with the Statemachine. I have already made these customizations locally and have also written UnitTests. The question is first of all whether the project is intended to make this information available.
If you are concerned about a condition that there are concrete indications, there is a risk for the application or security, it is free in the configuration of the condition to simply add no description for the error case.

If there is interest in the appropriate adjustments, I will gladly provide them in a branch.

nblumhardt:
Thanks for getting in touch!

If you're interested in submitting a PR and working through the process to get it merged, this sounds like a good improvement.

Could you share an example of the error messages before/after the proposed change?

HenningNT:
Is this functionality really needed? Will it clutter the API? I'm not sure, if we consider the Single Responsibility Principle, can this functionality be considered as part of the responsibility?
I think this feature is more a convenience, and it can be done by using a Trigger class object that contain the necessary information.

nblumhardt:
@HenningNT good questions; I think the before/after error messages would be a good illustration either way.

KrauseT:
Currently there are no error messages for transitions that can not be executed. Thus, it is not possible to display the error messages before expansion.
Currently there is only one public property for the transitions to be executed ("_PermittedTriggers_").
[...]

PR #160, opened 2017-04-10
Get descriptions for UnmetGuardConditions

KrauseT:
As promised in Issue #138 , I make the adjustments available for this topic. 
In addition, I have the idea with the additional out parameter included and implemented with. This idea makes it possible to get the error information for exactly one transition.
I hope that there will be a lively discussion about the adjustments and maybe parts of them will be adopted.

### DIFF
--- a/Stateless.nuspec
+++ b/Stateless.nuspec
@@ -1,0 +1,27 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<package xmlns="http://schemas.microsoft.com/packaging/2012/06/nuspec.xsd">
+  <metadata>
+    <id>Stateless-KrauseT</id>
+    <version>3.1.0-beta</version>
+    <title>Stateless-KrauseT</title>
+    <authors>Stateless Contributors</authors>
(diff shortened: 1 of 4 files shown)


nblumhardt:
@JeremyCade thanks for the nudge. I think this is an interesting idea, but needs more development to ensure it's possible to implement in a tight way.

In particular, the space of all triggers that _cannot_ be fired is effectively infinite in Stateless (let's imagine `TTrigger` is `int`. I'd want to be confident that, when the state machine surfaces information about not-permitted triggers, there's a meaningful boundary on that set, and not just circumstantial implementation decisions behind it. Does it matter that `4` is unavailable because of an unmet guard condition, vs. unavailable because the current state has no possible exit transition for `4`? As a particular state machine evolves, will this distinction seem arbitrary? (Not rhetorical questions - I haven't really dug deeply enough to decide for myself on this yet.)

Unless someone's keen to collect data on real-world use cases, I'd defer to @HenningNT 's intuitions on this for now.

HenningNT:

PR #168, opened 2017-05-19
Modified the output of OnUnmute to display the correct message

brianwhited:
The output message of the OnUnmute method displays the same message as the OnMute.


nblumhardt:
Cool, thanks for spotting this @brianwhited! :-)

Issue #167, opened 2017-05-18
Adding actions to Permit() and PermitIf(), and adding an OnExitFrom()


betty-crokker:
I've got some trial code in my fork: https://github.com/betty-crokker/stateless/tree/Permit

The best way to see the new APIs in action is in Tests/StateMachineFixture.cs.

There's a new (optional) triggerAction parameter on Permit(), looks like this:

```
sm.Configure(State.A)
    .Permit(Trigger.X, State.B, () => fired = true);
```
That's quite easy to use, and there's also an OnExitFrom() that gives the same functionality in a different way:

```
sm.Configure(State.A)
    .Permit(Trigger.X, State.B).OnExitFrom(Trigger.X, () => fired = true);
```
I also added the triggerAction parameter to OnPermitIf():

```
sm.Configure(State.A)
    .PermitIf(Trigger.X, State.B, () => true, () => value = 1)
    .PermitIf(Trigger.X, State.C, () => false, () => value = 2);

```
which I fear is a little hard to read, possible confusion as to which function is the guard and which is the action.  But OnExitFrom() doesn't help in this situation:

```
// Both trigger functions get called, regardless of the guard function
sm.Configure(State.A)
    .PermitIf(Trigger.X, State.B, () => true).OnExitFrom(Trigger.X, () => value = 1)
    .PermitIf(Trigger.X, State.C, () => false).OnExitFrom(Trigger.X,  () => value = 2);

```
Having an "OnExitFromIf" sounds really awkward.

If folks think this looks good I'll create the async versions ...

betty-crokker:
Per @nblumhardt  's suggestion, renamed OnExitFrom() to OnExitBy()

jsobell:
I suppose the question is whether this functionality belongs in the definition of the guards or at a central point.
Your solution addresses two issues; it allows a function to be specified, and it allows you to specify the data to pass to that function.
If I were using this I'd probably end up with every `PermitIf` calling `handleTransition(n)` where _n_ was a constant for each expression.
Since _n_ is simply an identifier for the appropriate transition, does it make more sense to have an ID with each transition, and allow an engine action passing that in as a parameter?
Your way is potentially less wasteful, as you can add or remove the 'event action' from each transition, but it's also far more verbose.
Also, if there are any other features required around transitions, we have the same issue again, in that we have no way to identify a transition. For example, we can currently query the trigger conditions, but not differentiate between two triggers with different guard conditions.
[...]

Issue #166, opened 2017-05-18
No way to identify the transition used to change states


jsobell:
At the moment we can have multiple PermitIf, so long as they are mutually exclusive, but I don't see any way to know which was in play when the guard was true.
In our state machine scenario we perform actions on transitions, so we either have to be able to add actions to a guarded transition (messy), or we have to use the `OnTransitioned` handler to look at the transition that was followed and act on that.
The issue at the moment is that there doesn't appear to be any way to know which transition was performed.  Sure, we can see the source and destination states, but when there are multiple transitions available between those states we don't know which it is.

Essentially, we need either an `OnTransition` raised by transitions, or the State's `OnEntry` has to be given something identifying the transition that went there. The current `OnTransitioned` is too late.
I would imagine we would also need an ID associated with the guarded transition. At the moment we have a description, but even that is not available.

Unless someone has another suggestion, how can an action be applied during the transitions below when we can't see which of these transitions was followed?

![image](https://cloud.githubusercontent.com/assets/834155/26182965/117c6902-3bbf-11e7-9afc-f0c25d66eed7.png)

HenningNT:
I would configure the OnButtonPress handler in the OffHook state to examine the number dialled, and log the number if needed, then post the CallDialed trigger. I don't really see the need to know which trigger caused a transition. In fact, I don't think a state should need to "know" anything about the past, but that might just be me.

jsobell:
I don't think any of that is relevant to this issue. You're suggesting every trigger call be prefixed with a check to see which guard would be met, and perform the action before the state machine is called? Doesn't that defeat the whole reason for using a state machine?

A state doesn't know anything about the past, but entering a state through a transition is not in the past, it's what is happening at that moment. `OnTransitioned` includes `source` and `destination`, and the transition is the path followed between them. The `trigger` is the thing that caused the transitions to be considered.

HenningNT:
No, that's not what I was suggesting.
Let's say that we have a phone class that has a few public methods, including ButtonPress(). Each time a button is pressed, the ButtonPress method is called. This method places an buttonPress event in the message queue. The phone object has a timer that checks the queue every so often, and calls Fire on the sttae machine. The state machine is configured to handle each button press in the OffHook state. When a valid number has been identified the CallDialed can be posted to the queue. The same handler can also determine if logging should be enabled or not.

jsobell:
[...]

PR #173, opened 2017-05-25
Add options to triggers

betty-crokker:
This code satisfies both issues #166  and #167 

By having Permit() return a TriggerConfiguration instead of a StateConfiguration (where TriggerConfiguration is a subclass of StateConfiguration), we make it easy to add options to the trigger using fluent syntax.

For example:

```
sm.Configure(State.A)
    .Permit(Trigger.X, State.B).If(anonymousGuard, "description");

sm.Configure(State.A)
    .Permit(Trigger.X, State.B).If(() => true).Do(() => value = 1)
    .Permit(Trigger.X, State.C).If(() => false).Do(() => value = 2);

sm.Configure(State.A)
    .Permit(Trigger.X, State.B).If(() => true).Tag(1)
    .Permit(Trigger.X, State.C).If(() => false).Tag("blue");
```

### DIFF
--- a/src/Stateless/StateConfiguration.Async.cs
+++ b/src/Stateless/StateConfiguration.Async.cs
@@ -24,8 +24,8 @@ namespace Stateless
             {
                 if (entryAction == null) throw new ArgumentNullException(nameof(entryAction));
 
-                _representation.AddTriggerBehaviour(new InternalTriggerBehaviour(trigger, guard));
-                _representation.AddInternalAction(trigger, (t, args) => entryAction(t));
+                Representation.AddTriggerBehaviour(new InternalTriggerBehaviour(trigger, guard));
+                Representation.AddInternalAction(trigger, (t, args) => entryAction(t));
                 return this;
             }
 
@@ -40,8 +40,8 @@ namespace Stateless
             {
                 if (internalAction == null) throw new ArgumentNullException(nameof(internalAction));
 
-                _representation.AddTriggerBehaviour(new InternalTriggerBehaviour(trigger, guard));
-                _representation.AddInternalAction(trigger, (t, args) => internalAction());
+                Representation.AddTriggerBehaviour(new InternalTriggerBehaviour(trigger, guard));
+                Representation.AddInternalAction(trigger, (t, args) => internalAction());
                 return this;
             }
 
@@ -57,8 +57,8 @@ namespace Stateless
             {
                 if (internalAction == null) throw new ArgumentNullException(nameof(internalAction));
 
-                _representation.AddTriggerBehaviour(new InternalTriggerBehaviour(trigger, guard));
-                _representation.AddInternalAction(trigger, (t, args) => internalAction(t));
+                Representation.AddTriggerBehaviour(new InternalTriggerBehaviour(trigger, guard));
+                Representation.AddInternalAction(trigger, (t, args) => internalAction(t));
                 return this;
             }
 
@@ -74,8 +74,8 @@ namespace Stateless
             {
                 if (internalAction == null) throw new ArgumentNullException(nameof(internalAction));
 
-                _representation.AddTriggerBehaviour(new InternalTriggerBehaviour(trigger.Trigger, guard));
-                _representation.AddInternalAction(trigger.Trigger, (t, args) => internalAction(ParameterConversion.Unpack<TArg0>(args, 0), t));
+                Representation.AddTriggerBehaviour(new InternalTriggerBehaviour(trigger.Trigger, guard));
+                Representation.AddInternalAction(trigger.Trigger, (t, args) => internalAction(ParameterConversion.Unpack<TArg0>(args, 0), t));
                 return this;
             }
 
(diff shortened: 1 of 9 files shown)


src/Stateless/StateConfiguration.Async.cs
```diff
@@ -24,8 +24,8 @@ public StateConfiguration InternalTransitionAsyncIf(TTrigger trigger, Func<bool>
             {
                 if (entryAction == null) throw new ArgumentNullException(nameof(entryAction));
 
-                _representation.AddTriggerBehaviour(new InternalTriggerBehaviour(trigger, guard));
-                _representation.AddInternalAction(trigger, (t, args) => entryAction(t));
+                Representation.AddTriggerBehaviour(new InternalTriggerBehaviour(trigger, guard));
```
HenningNT: Did you make Representation public? If not, could we keep this as _representation?

jsobell:
Nice. I've submitted a PR to your branch with the Tags being strongly typed too (but defaults to `object`)
Remains backwards compatible, but makes it a bit nicer than 'Object'.

src/Stateless/Transition.cs
```diff
@@ -45,6 +48,11 @@ public Transition(TState source, TState destination, TTrigger trigger)
             public TTrigger Trigger { get { return _trigger; } }
 
             /// <summary>
+            /// Optional data associated with the transition.
+            /// </summary>
+            public object Tag { get { return _tag; } }
```
HenningNT: In my opinion the tag object is not really needed.

jsobell: I think it is, for reasons in a huge thread on the issues :)

jsobell: If you support guards, you need to be able to differentiate between transitions, and this is the only practical way we can come up with.
What I think is not necessary is the `Do()` feature, as that can be replaced with user specific implementation of the `Tag`, but I have no issue leaving it in as a helper feature.

HenningNT: I don't like to have one parameter do more than one thing, it reduces the clarity and makes it harder for a user to understand the API. So if we want to associate an Action with a trigger I prefer to have an explicit way of doing so. So let's keep Do(Action).

jsobell: No, neither do I. The 'Do()' is effectively redundant, as we already have an event that fires when the transition occurs, and the only thing missing is how to associate the two.
This is an incredibly common scenario, and has dozens of ways of coding. If you considered 'Tag' to be metadata then you might declare its type as `<int, Action<>>` and simply call the action in the central handler. 
Here's the OnOffExample modified to provide an ID and a function in the definition:
```
using System;
using Stateless;

namespace OnOffExample
{
    class TransitionMetadata
    {
        public int Id;
        public Action Activity;

        public TransitionMetadata(int id, Action activity)
        {
            this.Id = id;
            this.Activity = activity;
        }
    }

    class Program
    {
        static void Main(string[] args)
        {
            try
            {
                string on = "On", off = "Off";
                var space = ' ';

                var onOffSwitch = new StateMachine<string, char, TransitionMetadata>(off);
                onOffSwitch.OnTransitioned(t =>
                {
                    Console.WriteLine($"Detected transition ID {t.Tag.Id}");
                    t.Tag.Activity?.Invoke();
                });
                onOffSwitch.Configure(off).Permit(space, on).Tag(new TransitionMetadata(100, () => Console.WriteLine("On was transitioned!")));
                onOffSwitch.Configure(on).Permit(space, off).Tag(new TransitionMetadata(200, () => Console.WriteLine("Off was transitioned!")));

                Console.WriteLine("Press <space> to toggle the switch. Any other key will raise an error.");

                while (true)
                {
                    Console.WriteLine("Switch is in state: " + onOffSwitch.State);
                    var pressed = Console.ReadKey(true).KeyChar;
                    onOffSwitch.Fire(pressed);
                }
            }
            catch (Exception ex)
            {
                Console.WriteLine("Exception: " + ex.Message);
                Console.WriteLine("Press any key to continue...");
                Console.ReadKey(true);
            }
        }
    }
}
```

Sure, it's just another way of doing things, but it's a single point of execution for transitions, which I think is much cleaner. It also removes ambiguity between the event handler and the Do() action, so you define the order yourself.
Personally, I don't like code in my workflows. Ideally I like them to be serialisable, which means no embedded lambdas or object references, and when we are auto-generating workflows from users' definitions I would rather add those to a switch statement than embed them as lamdas throughout the StateMachine configuration section.

betty-crokker:
Another benefit of having the options (e.g. "If" and "Do") in separate functions like that:

Right now we have 8 versions of OnEntryFrom(), representing (0 through 3 parameterized arguments) times (pass the transition as another argument yes or no).

If we follow that same pattern for PermitIf() and PermitDo() and PermitIfDo() we will have to add 6 versions of PermitIf, all 8 version of PermitDo, and 48 versions of PermitIfDo(), total of 62 new functions.

If instead we have Permit.If(), Permit.Do(), and Permit.If().Do(), we only have to write 14 new functions.

betty-crokker:
Do we feel strongly about the version of PermitIf() that takes an array of Tuples?
1) It's an awkward syntax to use (all those "new Tuple<Func<bool, string>")
2) You have to pass a string description (or empty string or null which looks funny)
3) I don't see how it gets us anything beyond just calling all of the possible guard functions with logical OR between them:

```
sm.Configure(State.B)
    .PermitIf(Trigger.X, State.A,
        new Tuple<Func<bool>, string>(GreenLightIsOn, "GreenLightIsOn"),
        new Tuple<Func<bool>, string>(RedLightIsOff, "RedLightIsOff"));
```

could just be

```
sm.Configure(State.B)
    .Permit(Trigger.X, State.A).If(() => GreenLightIsOn() | RedLightIsOff());
```

HenningNT:
One year later...
I think we should look into this again.
I would like to reduce the API, and the Permit.If().Do() syntax foes that. This would definitely be a breaking change, so it would go into stateless v5.
@betty-crokker would you like to update this PR sometime later, or maybe someone would like to do this?

HenningNT:

PR #175, opened 2017-05-31
Fix Multiple internal transition actions executed if defined in state hierarchy, and Renamed DynamicTransitionInfo.Destination to DestinationDescription

HenningNT:


nblumhardt:
Looks good, @HenningNT 👍

PR #178, opened 2017-06-08
Replace calls to Enforce.ArgumentNotNull with in-place null checks, and other changes suggested by CodeCracker

HenningNT:
Replaced calls to Enforce.ArgumentNotNull with in-place null checks
Removed unneccessary calls to ArgumentNotNull.
Inlined some parameter declarations.
Replaced explicit strings with nameof (parameterName) instead.
Removed dead method: InternalPermitDynamic(TTrigger trigger, Func<object[], TState> destinationStateSelector)
Added parameter element in documentation comments for TryGetMethodInfo and TryGetMethodName.


src/Stateless/Enforce.cs
```diff
@@ -7,13 +7,6 @@ namespace Stateless
 {
     static class Enforce
```
nblumhardt: Without any methods, can we just remove the class?

src/Stateless/ParameterConversion.cs
```diff
@@ -6,8 +6,6 @@ static class ParameterConversion
     {
         public static object Unpack(object[] args, Type argType, int index)
         {
-            Enforce.ArgumentNotNull(args, nameof(args));
```
nblumhardt: Missing a null-check here now?

src/Stateless/StateConfiguration.Async.cs
```diff
@@ -201,7 +201,6 @@ public StateConfiguration InternalTransitionAsync<TArg0>(TriggerWithParameters<T
             /// <returns>The receiver.</returns>
             public StateConfiguration OnActivateAsync(Func<Task> activateAction, string activateActionDescription = null)
             {
-                Enforce.ArgumentNotNull(activateAction, nameof(activateAction));
```
nblumhardt: In these public methods, keeping a null check seems worthwhile.

nblumhardt: Ah, I see - the check is performed in _representation.AddActivateAction(). Works for me! 👍

src/Stateless/ParameterConversion.cs
HenningNT: Looking up the call-chain, it looks like args can never be null, but it can be a zero length array. So in the current state it shouldn't require a null check. But I'll add one for future proofing.

src/Stateless/Enforce.cs
HenningNT: I had a weird error message after I deleted the class, the compiler could not find Enforce.Equals or something...
It doesn't complain now, though so I'll get rid of it.

nblumhardt:
👍

Issue #186, opened 2017-07-10
PermitDynamic does not handle reentry


tomthoros:
When using `PermitDynamic` with or without parameters it's impossible to stop it from reentering a state.

HenningNT:
I think we need a bit more information, could you provide an example / unit test that demonstrates this?

Does it always leave the state even when there is no valid destination state?

tomthoros:
> Does it always leave the state even when there is no valid destination state?

It correctly throws `No valid leaving transitions are permitted from state 'B' for trigger 'X'. Consider ignoring the trigger. ` exception if the StateSelector throws.

In the following program using `PermitDynamic` can reenter a state. But using `Permit` to try to reenter it throws an exception. I'm not sure what should happen actually, maybe this behavior is intended? But my first thought was that it wouldn't reenter a state.
The output of the following program is

```
Exit state A.
Enter state A.
Exit state A.
Enter state B.
Exit state B.
Enter state A.
```

```
    class Program
    {
        enum States 
        { 
            A,
            B,
        }
        enum Triggers
        {
            X,
            Y,
            Z
        }

        static States StateSelector(string parameter)
        {
            switch (parameter)
            {
                default: throw new ArgumentOutOfRangeException();
                case "A": return States.A;
                case "B": return States.B;
            }
        }

        static void Main(string[] args)
        {
            StateMachine<States, Triggers>.TriggerWithParameters<string> trig = new StateMachine<States,Triggers>.TriggerWithParameters<string>(Triggers.X);
            StateMachine<States, Triggers> sm = new StateMachine<States,Triggers>(States.A);
            sm.Configure(States.A)
                .OnEntry(() => { Debug.WriteLine("Enter state A."); })
                .OnExit(() => { Debug.WriteLine("Exit state A."); })
                .PermitDynamic(trig, StateSelector);
            sm.Configure(States.B)
                .OnEntry(() => { Debug.WriteLine("Enter state B."); })
                .OnExit(() => { Debug.WriteLine("Exit state B."); })
                //.Permit(Triggers.Y, States.B) // Illegal: Additional information: Permit() (and PermitIf()) require that the destination state is not equal to the source state. To accept a trigger without changing state, use either Ignore() or PermitReentry().
                .Permit(Triggers.Z, States.A);

            sm.Fire(trig, "A");
            sm.Fire(trig, "B");
            sm.Fire(Triggers.Z);

            Console.ReadLine();
        }
    }
```

[...]

PR #190, opened 2017-08-01
PermitDynamic does not handle reentry #186

tomthoros:
Add allowReentry parameter to PermitDynamic method overloads to specify whether the machine should reenter states or not.

### DIFF
--- a/src/Stateless/DynamicTriggerBehaviour.cs
+++ b/src/Stateless/DynamicTriggerBehaviour.cs
@@ -12,18 +12,22 @@ namespace Stateless
             readonly Func<object[], TState> _destination;
             internal Reflection.DynamicTransitionInfo TransitionInfo { get; private set; }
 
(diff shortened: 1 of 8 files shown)


nblumhardt:
I'm sorry about the slow response on this one, Tom. I meant to come back around to this, but have very limited time for this project at present. @dotnet-state-machine/reviewers is anyone able to check this one out? Thanks!

HenningNT:
I'm closing this as we discovered in #186 that it wasn't really a problem. Please let me know if I'm in error.

gerardog:
@HenningNT... I wouldn't say it's not a problem. It's just no longer a blocker for me as I stopped using `OnEntry` actions because of this issue, plus the difficulty that when an OnEntry action throws an exception, my external retry logic doesn't work (because it's based on retrying the last executed trigger, which will no longer trigger the correct OnEntry action, similar to what is posted on #267).

Issue #191, opened 2017-08-01
Shouldn't Reentry fire OnEntry from substate?


dpavlenkov:
`

        [Fact]
        public void IfSelfTransitionPermited_ActionsFire_InSubstate()
        {
            var sm = new StateMachine<State, Trigger>(State.A);

            bool fired = false;

            sm.Configure(State.B)
                .OnEntry(t => fired = true)
                .PermitReentry(Trigger.X);

            sm.Configure(State.A)
                .SubstateOf(State.B);

            sm.Fire(Trigger.X);

            Assert.Equal(State.B, sm.State); // Ok
            Assert.True(fired);              // Fail
        }
`

nblumhardt:
Thanks for the note; it appears to me that it should.

HenningNT:
Extending the test a bit further reveals that OnExit isn't called either...
        [Fact]
        public void IfSelfTransitionPermited_ActionsFire_InSubstate()
        {
            var sm = new StateMachine<State, Trigger>(State.A);

            bool onEntryStateBfired = false;
            bool onExitStateBfired = false;
            bool onExitStateAfired = false;

            sm.Configure(State.B)
                .OnEntry(t => onEntryStateBfired = true)
                .PermitReentry(Trigger.X)
                .OnExit(t => onExitStateBfired = true);

            sm.Configure(State.A)
                .SubstateOf(State.B)
                .OnExit(t => onExitStateAfired = true);

            sm.Fire(Trigger.X);

            Assert.Equal(State.B, sm.State); // Ok
            Assert.True(onEntryStateBfired);              // Fail
            Assert.True(onExitStateBfired);              // Fail
            Assert.True(onExitStateAfired);              // True
        }

HenningNT:
This also doesn't work properly:

        [Fact]
        public void FromSubstateToAnotherSubstae()
        {
            var sm = new StateMachine<State, Trigger>(State.A);

            bool onEntryStateDfired = false;
            bool onEntryStateCfired = false;
            bool onExitStateBfired = false;
            bool onExitStateAfired = false;

            sm.Configure(State.B)
                .OnExit(t => onExitStateBfired = true);

            sm.Configure(State.A)
                .SubstateOf(State.B)
                .Permit(Trigger.X, State.D)
                .OnExit(t => onExitStateAfired = true);

            sm.Configure(State.D)
                .OnEntry(t => onEntryStateDfired = true);

            sm.Configure(State.C)
[...]

PR #198, opened 2017-10-04
#191 Modified StateRepresentation.Exit to allow recursive Exit action…

HenningNT:
…s to be executed. It now modifies the Transition object as it executes Exit recursively. It returns a new transistion object when done, so that the source is set to the topmost superstate.

Modified StateMachine.InternalFireOne to accept the modified transition object returned by StateRepresentation.Exit.
Added needed test to verify correct handling of reeantrant transitions, as pointed out by @dpavlenkov

### DIFF
--- a/src/Stateless/StateRepresentation.cs
+++ b/src/Stateless/StateRepresentation.cs
@@ -176,7 +176,7 @@ namespace Stateless
                 }
             }
 
-            public void Exit(Transition transition)
+            public Transition Exit(Transition transition)
             {
                 if (transition.IsReentry)
                 {
@@ -189,8 +189,12 @@ namespace Stateless
                     ExecuteExitActions(transition);
 
                     if (_superstate != null)
-                        _superstate.Exit(transition);
+                    {
+                        transition = new Transition(_superstate.UnderlyingState, transition.Destination, transition.Trigger);
+                         return _superstate.Exit(transition);
+                    }
                 }
+                return transition;
             }
 
             void ExecuteEntryActions(Transition transition, object[] entryArgs)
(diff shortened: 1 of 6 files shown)


test/Stateless.Tests/StateMachineFixture.cs
```diff
@@ -442,5 +442,30 @@ public void IgnoreVsPermitReentryFrom()
 
             Assert.Equal(1, _numCalls);
         }
+        [Fact]
```
nblumhardt: Nit: whitespace

src/Stateless/StateRepresentation.cs
```diff
@@ -176,7 +176,7 @@ public void Enter(Transition transition, params object[] entryArgs)
                 }
             }
 
-            public void Exit(Transition transition)
+            public Transition Exit(Transition transition)
```
nblumhardt: Is there an `async` version of this method somewhere that we need to update?

HenningNT: There's 'public async Task ExitAsync(Transition transition)'. I guess this has to modified to return Task<Transition>...

HenningNT:
I've modified ExitAsync and InternalFireOneAsync as well.
To be honest, I don't like the async part of StateMachine at all. To make the test pass I added await sm.FireAsync(Trigger.X); Otherwise the Entry and Exit actions were executed on different Tasks that didn't complete before the Assert were executed.

Adding await to sm.FireAsync forces the state machine to become synchronous, which sort of defeats the purpose of making an async version in the first place :-D
I really think that actions should be run atomically and run-to-completion.

dpavlenkov:
Indeed, my usage of any control logic (which is what SM is) is to wrap it. If it wants to do something async, it has be an action to start something and trigger to complete it. I love Stateless API and idea, but I'm using my own implementation with just bare essentials, thoroughly tested, zero async.

nblumhardt:
@HenningNT Looks good! 👍 

> Adding await to sm.FireAsync forces the state machine to become synchronous

The thread will still be freed up to do other work while awaiting the call, which is the main aim; the async version of the state machine doesn't add any concurrency.

PR #213, opened 2017-11-16
Add support for trigger parameters inside the guard condition.

TheFo2sh:


dannyzhan:
This really solved my issue for guard with parameters, but can we use strong typed parameter instead of "object"?

TheFo2sh:
I think , we can but it need more time to get it working.

HenningNT:
@TheFo2sh This is great work! Would you mind following up on the suggestion fromo @dannyzhan, and also resolve the conflicts?

HenningNT:

PR #216, opened 2017-12-07
Fixed path in Stateless.sln and JsonExample project moved into example directory/.

tadassukys:
It was not able to load JsonExample project.


HenningNT:
Well done!
I should have checked this better ;-)

PR #221, opened 2018-01-15
IOworkflow Examples

GeXiaoguo:


stakx:
(Question from someone just passing by:) Shouldn't this be put in its own example project under `example`, instead of partially overwriting `BugTrackerExample`?

HenningNT:
HI, thanks for the PR!
This looks like a bit like work-in-progress, I can see alot of triggers defined, but not used.
Also I'd like to have more comments in the example code to explain people new to stateless what's going on. I also think that it should be in its own directory.

@GeXiaoguo, Would you like to expand on this example?

HenningNT:
Closing due to lack of feedback from @GeXiaoguo

Issue #209, opened 2017-11-03
Async implementation choices


smartcaveman:
I really like this library.  I like it so much that I've used it enough to run into frustrating situations with the async implementation with respect to composition and consistency:

1.  The async APIs do not support cancellation.  This limits how well Stateless can coordinate with external events in a system.

2.  The async APIs do not `ConfigureAwait`.  This limits the usefulness of Stateless state machines in some contexts because additional work is required to prevent deadlocks and unresponsive UIs.  

3.  `Fire` and `FireAsync` do not process events consistently with each other.  It's possible to run synchronous actions asynchronously (e.g. `Task.Run(...)`) and asynchronous actions synchronously (e.g. `task.Result`/`task.Wait()`).  Given that the formal semantics of an FSM do not consider concurrency, my initial expectation was for Stateless to coalesce the synchronous and asynchronous behavior.  Stateless will run synchronous callbacks asynchronously.  However, the converse call causes an error.
 So, it's impossible to synchronously fire a trigger with registered async handlers, and there's this counterintuitive situation where a client could run `if(machine.CanFire(trigger)){ machine.Fire(trigger); }` and get an `InvalidOperationException`.  

4.  At least this one is documented, but Stateless state machines are not thread-safe.  This limits how effectively Stateless can be leveraged in a system that demands concurrency.

I'm happy to assist with resolving any or all of these issues.  However, I'm aware that these issues could just as easily be caused by a design choice as by oversight.  So, my questions with respect to each issue are:

1.  Why does Stateless do it this way?
2.  Should Stateless continue to do it this way?

smartcaveman:
@HenningNT @nblumhardt

nblumhardt:
Howdy!

1. Cancellation support would be a great addition 👍 
2. Adding `ConfigureAwait()` would be a good bug-fix 👍 
3. This is a tough design space - `InvalidOperationException` seems like the best by-design option we have, for this one - I'm also okay with the asymmetry, the asymmetry is there in all C# that makes use of async
4. Stateless isn't designed to be used from multiple threads; it's not possible to make the current design thread-safe without the possibility of introducing deadlock foot-guns; concurrent state machine design is an interesting problem, might be better explored in a new library

HTH, and thanks for the input! :-)

smartcaveman:
@nblumhardt - re #3, what do you think of a configuration setting

smartcaveman:
I can help with #1 and #2 if there are no one else is already working on it.

[...]

PR #222, opened 2018-01-15
Added ConfigureAwait(false) calls

pdinnissen:
I do not believe that context matters, so ConfigureAwait(false) should be added to all await calls. #209


twsouthwick:
@pdinnissen There is a [set of analyzers](https://www.nuget.org/packages/Microsoft.CodeAnalysis.FxCopAnalyzers) to replace FxCop that includes an analyzer to ensure that nothing is awaited without `ConfigureAwait()`. May be helpful to ensure 1) you got all the instances, and 2) catch any instances in the future.

HenningNT:
Looks good, thanks alot!

PR #235, opened 2018-03-09
Fix Issue 228, Guard Conditions being evaluated more than once.

trgrote:
Simple fix to ensure guard conditions are only evaluated once during a transition.


HenningNT:
Looks good, thank you!

Issue #214, opened 2017-11-20
Transition object holds parent state as source state instead of child


xpleria:
Stateless.StateMachine<State, Trigger> sm = new Stateless.StateMachine<State, Trigger>(State.GrandParent);

            Stateless.StateMachine<State, Trigger>.Transition t = null;

            sm.Configure(State.GrandParent)
                .OnEntry(transition => { t = transition; })
                .Permit(Trigger.GtoP, State.Parent);

            sm.Configure(State.Parent)
                .SubstateOf(State.GrandParent)
                .OnEntry(transition => { t = transition; })
                .Permit(Trigger.PtoA, State.Aunt)
                .Permit(Trigger.PtoCA, State.ChildA);

            sm.Configure(State.ChildA)
                .SubstateOf(State.Parent)
                .OnEntry(transition => { t = transition; })
                .Permit(Trigger.CAtoCB, State.ChildB);

            sm.Configure(State.ChildB)
                .SubstateOf(State.Parent)
                .OnEntry(transition => { t = transition; })
                .Permit(Trigger.CBtoCA, State.ChildA);

            sm.Configure(State.Aunt)
                .SubstateOf(State.GrandParent)
                .OnEntry(transition => { t = transition; })
                .Permit(Trigger.AtoP, State.Parent);

            sm.Fire(Trigger.GtoP);
            Console.WriteLine("From: {0} To: {1} Trigger: {2}", t.Source, t.Destination, t.Trigger);

            sm.Fire(Trigger.PtoCA);
            Console.WriteLine("From: {0} To: {1} Trigger: {2}", t.Source, t.Destination, t.Trigger);

            sm.Fire(Trigger.PtoA);
            Console.WriteLine("From: {0} To: {1} Trigger: {2}", t.Source, t.Destination, t.Trigger);

            sm.Fire(Trigger.AtoP);
            Console.WriteLine("From: {0} To: {1} Trigger: {2}", t.Source, t.Destination, t.Trigger);

            sm.Fire(Trigger.PtoCA);
            Console.WriteLine("From: {0} To: {1} Trigger: {2}", t.Source, t.Destination, t.Trigger);

            sm.Fire(Trigger.CAtoCB);
            Console.WriteLine("From: {0} To: {1} Trigger: {2}", t.Source, t.Destination, t.Trigger);

            sm.Fire(Trigger.PtoA);
[...]

PR #245, opened 2018-04-19
Fixing #214 by invoking the OnTransition action with the original sou…

HenningNT:
…rce and destination.


HenningNT:
I'll merge this tomorrow if there is no protests.

Issue #223, opened 2018-01-15
Queuing should not be the default behavior


khaledsobhy83:
I was using stateless library before introducing the queuing part and everything was just fine with me. Now after upgrading to the latest .Net Standard version, the queuing of triggers cause my application to totally fail. The point is that previously when calling .Fire I was expecting the event to be triggered but now it will be queued if the parent is still being processed and will not be executed. Assume a scenario when having two states Processing and checkpointing every one of these states has a begin and finish triggers, the beginprocessing of the first state fires the begincheckpointing of the second state then fires its finishprocessing event, assuming that when begincheckpoint is fired, it will call the finishcheckpoint first then the finishprocessing will be triggered. With the introduction of queuing part what will happen is that begincheckpoint, finishprocessing will be added to the queue in the wrong order, the correct order should be begincheckpoint, finishcheckpoint then finishprocessing.

I think the queuing is a nice feature but it should not be the default behavior, there should be a way to enforce the execution when calling .Fire like before.

HenningNT:
I tend to disagree. Run-to-completion is important to guarantee atomic execution of events and state transitions. Therefore, allowing Fire to execute while processing a trigger should not be allowed.

khaledsobhy83:
Exactly, this is what I want, Run-to-Completion, but what is currently happening is that instead of running and executing the trigger, they will be queued which does not mean actual execution and using the queue will end up of wrong order of items in the queue. Again here is the scenario and please tell me how to handle it
private void BeginProcessRecords()
        {
            _stateMachine.Fire(Trigger.BeginCheckPoint); // This calls FinishCheckpoint
            _stateMachine.Fire(Trigger.FinishProcessRecords); // This cannot be triggered before FinishCheckPoint is triggered
        }

What will happen is that During the execution of BeginProcessRecords, the BeginCheckPoint and FinishProcessRecords will be queues which is completely wrong, I want FinishCheckPoint to be executed before FinishProcessRecords.
Can you tell me how to guarantee that FinishProcessRecords will be triggered after FinishCheckpoint?

HenningNT:
We'll, I "solved" my problem by posting all events to an external queue. I don't call Fire at all onside the state machine, I have timer that empties the queue at regular intervals. This allows me to also post lifo if needed.

khaledsobhy83:
I added another method that does not use the queue and did a commit. not sure how the review process works. It will be another optional way to fire events the old way before the queuing part.

[...]

PR #246, opened 2018-04-20
#223 making queueing optional

HenningNT:


src/Stateless/StateMachine.cs
```diff
@@ -5,6 +5,16 @@
...
+    /// Enum for the different modes used when Fire-ing a trigger
+    /// </summary>
+    public enum FiringMode
+    {
+        /// <summary> Use immediate mode when the queing of trigger events are not needed. Care must be taken when using this mode, as there is no run-to-completion guaranteed.</summary>
+        Immediate,
+        /// <summary> Use the queued Fire-ing mode when run-to-completion is required. This is the recommended mode.</summary>
+        Queued
+    }
     /// <summary>
```
nblumhardt: Nitpick: missing whitespace :-)

src/Stateless/StateMachine.cs
```diff
@@ -18,6 +28,7 @@ public partial class StateMachine<TState, TTrigger>
         readonly Action<TState> _stateMutator;
         UnhandledTriggerAction _unhandledTriggerAction;
         OnTransitionedEvent _onTransitionedEvent;
+        private readonly FiringMode _firingMode;
```
nblumhardt: Nitpick: explicit `private` is inconsistent with the lines above.

src/Stateless/StateMachine.cs
```diff
@@ -5,6 +5,16 @@
...
 {
+    /// <summary>
+    /// Enum for the different modes used when Fire-ing a trigger
+    /// </summary>
+    public enum FiringMode
+    {
+        /// <summary> Use immediate mode when the queing of trigger events are not needed. Care must be taken when using this mode, as there is no run-to-completion guaranteed.</summary>
+        Immediate,
+        /// <summary> Use the queued Fire-ing mode when run-to-completion is required. This is the recommended mode.</summary>
+        Queued
```
nblumhardt: Although a queue is used for the implementation, would it be clearer to call this something like `OnCompletedTransition`?

src/Stateless/StateMachine.cs
```diff
@@ -33,21 +44,25 @@ class QueuedTrigger
         /// </summary>
         /// <param name="stateAccessor">A function that will be called to read the current state value.</param>
         /// <param name="stateMutator">An action that will be called to write new state values.</param>
-        public StateMachine(Func<TState> stateAccessor, Action<TState> stateMutator) : this()
+        /// <param name="firingMode">Optional specification of fireing mode.</param>
+        public StateMachine(Func<TState> stateAccessor, Action<TState> stateMutator, FiringMode firingMode = FiringMode.Queued) : this()
```
nblumhardt: Does this need to be a binary breaking change, or could an additional overload be added? For cases where this package is a dependency-of-a-dependency, the change in constructor signature could cause problems.

src/Stateless/StateMachine.cs
```diff
@@ -33,21 +44,25 @@ class QueuedTrigger
...
         /// </summary>
         /// <param name="initialState">The initial state.</param>
-        public StateMachine(TState initialState) : this()
+        /// <param name="firingMode">Optional specification of fireing mode.</param>
+        public StateMachine(TState initialState, FiringMode firingMode = FiringMode.Queued) : this()
         {
             var reference = new StateReference { State = initialState };
             _stateAccessor = () => reference.State;
             _stateMutator = s => reference.State = s;
+            _firingMode = firingMode;
```
nblumhardt: If we check that the firing mode is either `Immediate` or `Queued` here, we can get rid of the switch and default case below? (Just a thought - may not really be a big advantage).

src/Stateless/StateMachine.cs
HenningNT: I could add another constructor, but which mode should the current constructor use?

Would the dependency-of-a-dependency issue really be an issue, since the dependency is always (?) on a specific version?

src/Stateless/StateMachine.cs
HenningNT: It is certainly possible to add an Action<TTrigger, object[]> FireHandler that is configured in the constructor.  InternalFire is then reduced to 

        {
            FireHandler(trigger, args);
        }

It would shave off a few CPU cycles...

src/Stateless/StateMachine.cs
HenningNT: I'll add private to everything :-D

src/Stateless/StateMachine.cs
HenningNT: I'm on it!

src/Stateless/StateMachine.cs
nblumhardt: :+1:

HenningNT:
If there are no further comments I'll merge this tomorrow :godmode:

src/Stateless/StateMachine.cs
nblumhardt: The current constructor should use queued, since it's the existing default, correct?

The problem regarding the dependency is when package A depends on the old version, B depends on the new version, and a user wants to use both A and B. With this change, that scenario is impossible, unfortunately.

nblumhardt:
Just a note regarding the package version, if the old constructor is kept for compatibility, this changeset needs a version bump to `4.1` (added features). If you decide to go with the new constructor only, the version bump needs to be to `5.0` (breaking changes). (Current dev version is at `4.0.1`).

src/Stateless/StateMachine.cs
HenningNT: So what would be the best approach? Keep the original constructor(s), and add another set of constructors With the fireing mode?

HenningNT:
I think I'd like to only go to 4.1 for now 👍

src/Stateless/StateMachine.cs
nblumhardt: It's a pretty quick transformation from where the PR is now, I think -

 * The extended one (that accepts firing mode) loses the default value (otherwise overload resolution fails for callers)
 * Another constructor is added without the parameter, which just calls the new, longer constructor, passing in the "queued" default value

If you need a hand I can jump in and edit, just give the word :-) Cheers!

HenningNT: I think I understood it, take a look 😁

src/Stateless/Properties/AssemblyInfo.cs
```diff
@@ -8,7 +8,7 @@
 [assembly: AssemblyCopyright("Copyright © Stateless Contributors 2009-2016")]
 
 // This value will remain fixed throughout the 4.x series.
-[assembly: AssemblyVersion("4.0.0.0")]
+[assembly: AssemblyVersion("4.1.0.0")]
```
nblumhardt: This one shouldn't be updated, just the CSPROJ

nblumhardt:
LGTM! Just had the one comment on assembly info.

:shipit:

src/Stateless/Properties/AssemblyInfo.cs
HenningNT: Ok, I'll sort it out.

PR #250, opened 2018-05-03
#223 Adding support for Async firing mode as well.

HenningNT:
This duplicates the code in the Sync version.
Also had to change to switch..case to be able to use the same constructor for all combinations.


HenningNT:
Since this is not a very controversial change that only completes a feature I'll merge this tomorrow if there are no objections.

PR #251, opened 2018-05-04
Adding check in InternalTransitionAsync*, checking if supplied method…

HenningNT:
Checks added to InternalTransitionAsync***, to guard against assigning a Sync method to the internal action handler. This would throw an null reference exception when running FireAsync.

### DIFF
--- a/src/Stateless/StateConfiguration.Async.cs
+++ b/src/Stateless/StateConfiguration.Async.cs
@@ -4,6 +4,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
+using System.Runtime.CompilerServices;
 using System.Text;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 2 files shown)


FantasyTeddy:
Hi @HenningNT 

I just stumbled across this changeset because I wanted to upgrade from Stateless version 4.0.0 to 4.2.1 and I had to rewrite all the `InternalTransitionAsync()` calls from something like this:
```c#
.InternalTransitionAsync(() => SomeMethodReturningATask())
```
to this:
```c#
.InternalTransitionAsync(async () => await SomeMethodReturningATask())
```

Can you elaborate why you had to introduce the check for the `async` keyword?
In my understanding it should not matter if I pass in an `async` function or simply pass the corresponding `Task`. The resulting call to the `Func<Task>` should work anyways.

Thanks for explaining.

HenningNT:
Good question!
I think that if it worked in 4.0.0, then it should continue to work in all v4.x.x...
This should work in the same way as OnEntryAsync and needs to reworked.

HenningNT:
@FantasyTeddy If you have some time available please feel free to contribute, I'm out of time right now :-(

FantasyTeddy:
I would love to contribute, but unfortunately I don't think that I know your code base well enough to make clever changes.
My suggestion would be to undo this pull request, but I guess this is here for a reason...
If you have any hints to what breaks or what changes would be necessary, I'm more than willing to help.

Issue #249, opened 2018-04-26
Latest prerelease v4.0.1-dev-00294 bug


IanElsinga:
I have been using Stateless Nuget stable version 4.0.0 but ran into the InternalTransitionIf bug #227  so I attempted to use the above Prerelease but that is throwing an exception with a simple async trigger fire that was never thrown before in an unrelated place in the state machine.   (Unrelated meaning it is transitioning to a state that does not use InternalTransitionIf)

`await sm.FireAsync(Trigger.Approved);`

Causes the following exception: 
```   
at Stateless.StateMachine`2.StateRepresentation.<ExecuteInternalActionsAsync>d__13.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Stateless.StateMachine`2.<InternalFireOneAsync>d__18.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Stateless.StateMachine`2.<InternalFireAsync>d__17.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.GetResult()
   at LocalNamespaceNameRemoved.<OnDriverTransactionResponse>d__159.MoveNext() in …\DriverTransactionResponse.cs:line 66
```
Thanks, please let me know if it looks like this will take a while to fix and I'll rewrite to avoid InternalTransitionIf and go back to v4.0.0

HenningNT:
@IanElsinga would you mind taking a look at the 299 prerelease?
Could I salso ask you to post a test that demonstrate the problem?

HenningNT:
That's OK, I used the test in #227 and changed to FireAsync. It fails :-(
I'll take  a look...

HenningNT:
[...]

PR #252, opened 2018-05-10
#249 fix async internal transition not firing when sync

HenningNT:
Fixed by not casting to Async, both async and sync versions can call ExecuteAsync.


HenningNT:
Merging this so that we can get a new pre-release for testing.

Issue #228, opened 2018-02-10
guard function called twice?


Avareto:
Hi!
Can someone explain, why the guard function is called two times, when firing goNext? I was expecting it should be called only once. Here is my code (or have look at test function "WhenDiscriminatedByGuard_ChoosesPermitedTransition()" in StateMachineFixture.css, line 164 in the repro)
`enum S { State1, State2 };
enum T { goNext }

[TestMethod]
public void TestMethod1()
{
        var currentState = S.State1;
        var sm = new StateMachine<S, T>(() => currentState, s => currentState = s);
        sm.Configure(S.State1)
                .PermitIf(T.goNext, S.State2, () => guard());
        sm.Fire(T.goNext);
}
// called twice...???
public bool guard()
{
        return true;
}
`

HenningNT:
It shouldn't be called twice, which version are you using?

Avareto:
From Nuget: v4.0.0

HenningNT:
ok, that's a bit odd... I'll take a look.

HenningNT:
I had a look and that is actually by design!

There is a PermittedTriggers property, which will evaluate all guard functions to figure out which trigger can be executed. The state machine will use this when firing triggers.

Avareto:
Unfortunately, I have to do some expensive database operations to evaluate the guard status. This behavior will probably slow down my system significantly. 
Any way to get around it?

Avareto:
Sorry, but its much worse than I expected: When I use the same guard function in more than one PermitIf(…) statements, in my case in 3 state configurations, and I am triggering just one of those transitions, the guard function is called 6 times just for that transition. Can you please explain why it’s called 6 times and why this is by design? There should be a way to change the behavior? Otherwise the version of the library is unusable for me, which is unfortunate because I like it a lot?

HenningNT:
Well, saying by design was maybe a bit incorrect, I wasn't involved when that bit of code was written.

I think you have a few options:
1) Add a time stamp to the guard function, and cache the result. Invalidate the result if more than 260ms have passed since the last call to the guard.
2) Your state machine seems to depend on some external state. Model the dependency as another state machine. Then the new one can either be polled, or send messages to the first one about state changes.
3) Change stateless code so that the guard functions are called only once per trigger.

I'm hoping you'll do number 3 so everybody can enjoy the improvement 😁

trgrote:
[...]

PR #255, opened 2018-05-14
Release v4.1

HenningNT:
New features:
- Trigger queuing is now optional (#223)

Bugs fixed:
- #191 Shouldn't Reentry fire OnEntry from substate?
- #228 guard function called twice? 
- #214 Transition object holds parent state as source state instead of child
- #249 Latest prerelease v4.0.1-dev-00294 bug (Internal transition handler not executing)

Other:
- Added JSON example
- Added ConfigureAwait(false)


HenningNT:
@dotnet-state-machine/reviewers Please review and comment.
If no objections I'll merge this on Friday.

@nblumhardt I might need help with the build script...

HenningNT:
Looks like the version suffix created here...

    $suffix = @{ $true = ""; $false = "$($branch.Substring(0, [math]::Min(10,$branch.Length)))-$revision"}[$branch -eq "master" -and $revision -ne "local"]

... is null when branch is master?

nblumhardt:
Should be fixed now; it was a regression between v1 and v2 of the .NET Core tooling.

Issue #275, opened 2018-06-28
Accessing StateMachine.PermittedTriggers throws System.ArgumentException


rose-a:
Accessing the `StateMachine.PermittedTriggers` property throws a `System.ArgumentException` with the message `An argument of type System.String is required in position 0.`

StackTrace:
```
   bei Stateless.ParameterConversion.Unpack(Object[] args, Type argType, Int32 index)
   bei Stateless.ParameterConversion.Unpack[TArg](Object[] args, Int32 index)
   bei Stateless.StateMachine`2.TransitionGuard.<>c__DisplayClass4_0`1.<ToPackedGuard>b__0(Object[] args)
   bei Stateless.StateMachine`2.TransitionGuard.<>c__DisplayClass17_0.<UnmetGuardConditions>b__0(GuardCondition c)
   bei System.Linq.Enumerable.WhereSelectListIterator`2.MoveNext()
   bei System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   bei System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   bei Stateless.StateMachine`2.TransitionGuard.UnmetGuardConditions(Object[] args)
   bei Stateless.StateMachine`2.StateRepresentation.<>c__DisplayClass69_0.<GetPermittedTriggers>b__2(TriggerBehaviour a)
   bei System.Linq.Enumerable.Any[TSource](IEnumerable`1 source, Func`2 predicate)
   bei Stateless.StateMachine`2.StateRepresentation.<>c__DisplayClass69_0.<GetPermittedTriggers>b__0(KeyValuePair`2 t)
   bei System.Linq.Enumerable.WhereSelectEnumerableIterator`2.MoveNext()
   bei System.Linq.Buffer`1..ctor(IEnumerable`1 source)
   bei System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
   bei Stateless.StateMachine`2.StateRepresentation.GetPermittedTriggers(Object[] args)
   bei Stateless.StateMachine`2.get_PermittedTriggers()
```

kpi6288:
I can verify this for the project TelephoneCallExample. Appearently the last version that does not show this exception is the commit dated Mar 24, 2018.

smartcaveman:
[...]

Issue #272, opened 2018-06-15
Issue with InternalTransitionIf and trigger with single parameter


matt-w-cotton:
First off, thank you for a terrific library.
I have yet to track down the exact cause of the issue, but I assume it popped up when adding the `Func<object[], bool>` InternalTransitionIf guards.

Basically now when you attempt to `GetPermittedTriggers()`, if you have InternalTransitionIf, using trigger with single param, you will get:
`System.ArgumentException : An argument of type System.String is required in position 0.`

I added a quick test to demonstrate the issue.  Same test with 2 params and the issue does not come up.  `Unpack` is never called and thus never throws.
Here is the example:
```[Fact]
public void ConditionalInternalTransitionWithOneParam_ShouldWork()
{
    var isPermitted = true;
    var sm = new StateMachine<State, Trigger>(State.A);

    var trig = sm.SetTriggerParameters<string>(Trigger.X);
    sm.Configure(State.A)
        .InternalTransitionIf(trig, u => isPermitted, (u, t) => { });

    Assert.Equal(1, sm.GetPermittedTriggers().ToArray().Length);
    isPermitted = false;
    Assert.Equal(0, sm.GetPermittedTriggers().ToArray().Length);
}
```

and stack trace:
```System.ArgumentException : An argument of type System.String is required in position 0.
Stack Trace:
C:\dev\main\stateless\src\Stateless\ParameterConversion.cs(15,0): at Stateless.ParameterConversion.Unpack(Object[] args, Type argType, Int32 index)
C:\dev\main\stateless\src\Stateless\ParameterConversion.cs(26,0): at Stateless.ParameterConversion.Unpack[TArg](Object[] args, Int32 index)
C:\dev\main\stateless\src\Stateless\TransitionGuard.cs(19,0): at Stateless.StateMachine`2.TransitionGuard.<>c__DisplayClass4_0`1.<ToPackedGuard>b__0(Object[] args)
C:\dev\main\stateless\src\Stateless\TransitionGuard.cs(111,0): at Stateless.StateMachine`2.TransitionGuard.<>c__DisplayClass17_0.<UnmetGuardConditions>b__0(GuardCondition c)
at System.Linq.Enumerable.WhereSelectListIterator`2.MoveNext()
at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)          
at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
[...]

PR #276, opened 2018-07-10
Hotfix/can fire trigger with parameters

smartcaveman:
Implemented a minimal patch to address bugs with StateMachine.CanFire and StateMachine.PermittedTriggers under certain conditions,  

- Related Issues:
  - https://github.com/dotnet-state-machine/stateless/issues/272
  - https://github.com/dotnet-state-machine/stateless/issues/275

Notes:
- StateMachine.CanFire exhibits unexpected and inconsistent behavior when (1) its argument is a trigger with parameters and (2) an internal transition is registered.
- Added additional test cases to demonstrate scope of the problem.  Rather than simply affecting internal transitions, all transition configurations that depend upon conditional guard clauses are impacted.  The reason that non-conditional internal transition configurations are affected is that their implementation delegates to the conditional method with a condition that is always true.
- Fix for `StateMachine.CanFire`: The problematic validation fails on parameter length.  CanFire initializes an array of the appropriate length before invoking the underlying StateRepresentation method
- Fix for `StateMachine.PermittedTriggers`/`StateMachine.ToString()`: Changed the implementation of the StateMachine.PermittedTriggers property.  This resolves the exception that was being thrown.  It does not address StateMachine.GetPermittedTriggers(object[]) or any issues with the underlying StateRepresentation class.  StateMachine.ToString() depends on GetPermittedTriggers(), which remains either broken or unspecified in some cases.  It will still throw unexpected errors for triggers with parameters, which means that StateMachine.ToString() would do the same.  The assumption justifying this commit is that the reduced consistency between StateMachine.PermittedTriggers and StateMachine.GetPermittedTriggers() is less harmful than StateMachine.CanFire() and StateMachine.ToString() violating their contracts.

### DIFF
--- a/src/Stateless/StateMachine.cs
+++ b/src/Stateless/StateMachine.cs
@@ -494,7 +494,7 @@ namespace Stateless
         /// <returns>True if the trigger can be fired, false otherwise.</returns>
         public bool CanFire(TTrigger trigger)
         {
-            return CurrentRepresentation.CanHandle(trigger);
+            return CurrentRepresentation.CanHandle(trigger, new object[_triggerConfiguration.TryGetValue(trigger, out var triggerConfiguration) ? triggerConfiguration.Arity : 0]);
         }
 
         /// <summary>
--- a/src/Stateless/StateMachine.cs
+++ b/src/Stateless/StateMachine.cs
@@ -117,7 +117,7 @@ namespace Stateless
         {
             get
             {
-                return GetPermittedTriggers();
+                return CurrentRepresentation.TriggerBehaviours.Keys;
             }
         }
 
@@ -506,7 +506,7 @@ namespace Stateless
             return string.Format(
                 "StateMachine {{ State = {0}, PermittedTriggers = {{ {1} }}}}",
                 State,
-                string.Join(", ", GetPermittedTriggers().Select(t => t.ToString()).ToArray()));
+                string.Join(", ", PermittedTriggers.Select(t => t.ToString()).ToArray()));
         }
 
         /// <summary>
--- a/src/Stateless/StateMachine.cs
+++ b/src/Stateless/StateMachine.cs
@@ -116,8 +116,11 @@ namespace Stateless
         public IEnumerable<TTrigger> PermittedTriggers
         {
             get
-            {
-                return CurrentRepresentation.TriggerBehaviours.Keys;
+            { 
+                var set = new HashSet<TTrigger>();
+                for (var state = CurrentRepresentation; state != null; state = state.Superstate)
+                    set.UnionWith(state.TriggerBehaviours.Keys);
+                return set; 
             }
         }
 
--- a/src/Stateless/TriggerWithParameters.cs
+++ b/src/Stateless/TriggerWithParameters.cs
@@ -29,6 +29,17 @@ namespace Stateless
             public TTrigger Trigger { get { return _underlyingTrigger; } }
 
             /// <summary>
+            /// Gets the number of parameters associated with the trigger
+            /// </summary>
+            internal virtual int Arity => this.GetType()
+#if NET40
+                .GetGenericArguments()
+#else
+                .GenericTypeArguments
+#endif
+                .Length;
+
+            /// <summary>
             /// Ensure that the supplied arguments are compatible with those configured for this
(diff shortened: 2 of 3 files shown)


smartcaveman:
@HenningNT @nblumhardt - 
What needs to happen to get this PR merged?  I have a project going into production that needs it and I'd like to avoid creating an alternative NuGet off my fork, if possible. Please advise. Thank you

src/Stateless/TriggerWithParameters.cs
```diff
@@ -28,6 +28,17 @@ public TriggerWithParameters(TTrigger underlyingTrigger, params Type[] argumentT
             /// </summary>
             public TTrigger Trigger { get { return _underlyingTrigger; } }
 
+            /// <summary>
+            /// Gets the number of parameters associated with the trigger
+            /// </summary>
+            internal virtual int Arity => this.GetType()
```
nblumhardt: If this was defined `... Arity { get; } = ...` then initialization would occur only once, rather than every time the property is accessed.

src/Stateless/StateMachine.cs
```diff
@@ -494,7 +497,7 @@ public bool IsInState(TState state)
         /// <returns>True if the trigger can be fired, false otherwise.</returns>
         public bool CanFire(TTrigger trigger)
         {
-            return CurrentRepresentation.CanHandle(trigger);
+            return CurrentRepresentation.CanHandle(trigger, new object[_triggerConfiguration.TryGetValue(trigger, out var triggerConfiguration) ? triggerConfiguration.Arity : 0]);
```
nblumhardt: Any way to avoid the extra allocation here? (Declaring a static array of stand-in argument arrays, and then indexing into it by `Arity`, could do it - but with that level of machinery then perhaps `TriggerWithParameters` could just initialize and store a surrogate argument array and provide that directly instead of `Arity`?)

HenningNT:
<Rant>The root of the problem is that we try to evaluate the guards to get the actionable triggers. I think this is wrong. If we could I would have changed the code so that the user only gets a list of triggers allowed for a particular state, which didn't try to evaluate the guards. The state of a state machine is transient, so a list of permitted triggers is also transient in nature. Running the guards could have side effects, affecting the state of the state machine, and the state of other resources.
Changing the behaviour now would certainly break functionality for out consumers, so we are stuck with what we have. </Rant>

@nblumhardt had some performance related comments. I suggest revising the PR with his suggestions. If you don't want to do that then I guess we could push it out as it is. @JeremyCade is the owner of stateless, and he might have an opinion about this.

@smartcaveman I would appreciate it if you could update the change log as well.

src/Stateless/StateMachine.cs
smartcaveman: @nblumhardt I considered that, but this actually does less work because, in practice, `Arity` is overridden by all of its implementors to return a constant value and the reflection based method would only ever be called if someone were to add additional implementations in a client library.  The reason this seemed better than just leaving it abstract is that `Arity` can remain internal but still be compatible with a client library that follows the same pattern/conventions but used more parameters than are provided in Stateless.

That being said, I don't think the difference is that significant so it can certainly be changed as you suggest.  Do you still think adding a static array is the way to go here?

smartcaveman:
@henningNT happy to follow up on this.  I ended up replacing Stateless with a smaller custom component on the project I referenced, so that it should be less time sensitive for me. I left some comments in reply to @nblumhardt suggestions above, so I'll wait for that reply before implementing further changes.

I just want to point out that , with respect to the risk of this code "break functionality for our consumers": the current state of things is that there are cases where the `ToString()` method of the main entry point to the library throws an exception.

smartcaveman:
@HenningNT @nblumhardt @JeremyCade  still waiting for further guidance here ...

bravecobra:
Any progress on this PR?

HenningNT:
I'm thinking we can merge this as it is. Unless someone objects?

HenningNT:
I found that Unpack threw an exception when there where no parameters. This should be OK, so I
ve added a check for this in the Unpack method. I'll decline this PR, but I've added the unit tests to the repository. Sorry about this dragging out so long, thank so much for helping out!

Issue #189, opened 2017-07-28
Support async conditions in PermitIf


alexwilsh:
Hi,
It's possible I'm not fully understand the whole idea since I just started using Stateless state machine, but I'm curious why there is no PermitIfAsync method.
I'm trying to implement workflow using Stateless. So, for example, I have a document, which should have a workflow with certain states:
```
enum DocumentState
{
	New,
        Processing,
	Completed
};
```
That's clear that there are triggers for those states: StartProcessing, Complete.
All I want is to create async actions to support my workflow. In order to do that, I created action's interface:
```
public interface IWorkflowAction<TEntity>
	where TEntity : class
{
	Task<bool> CanExecuteAsync(TEntity entity);
	Task ExecuteAsync(TEntity entity);
}
```
I can't rely only on possible status transitions. I want also to call some external logic to check if the transition is possible. Like for example:
```
DocumentWorkflowAction startProcessingAction;
stateMachine.Configure(DocumentState.New)
	.PermitIf(DocumentTrigger.StartProcessing, DocumentState.Processing, 
        () => await startProcessingAction.CanExecuteAsync(m_document));

stateMachine.Configure(DocumentState.Processing)
	.OnEntryAsync(async () => await startProcessingAction.ExecuteAsync(m_document));
```
As you see, I'd like to call CanExecuteAsync for my document entity to check for example if an user has permissions to start processing this document. This operation has to be asynchronous since it can require for instance database query to be executed.
Following this logic, if an user doesn't have permissions, the transition is not allowed. If yes, then we call ExecuteAsync method on entering the Processing state to start processing a document.

Actually, I can't do that because PermitIf expects synchronous delegate.

I probably can call my CanExecuteAsync method on exit from New state or before calling ExecuteAsync method inside OnEntry of Processing state configuration, but I also need to block state transitioning so everyone should know that this transition is not possible now.

Please tell me whether this possible to resolve my task with existing functionality or not. If not, is there a reason not to add PermitIfAsync method with asynchronous delegate?

nblumhardt:
Hi Alex, thanks for the note. It looks like the absence of `PermitIfAsync()` is just an oversight in the current Stateless API. I don't know how much internal churn would be required in order to add it - needs some investigation. Cheers!

alexwilsh:
Nick,
Thanks for the response. At least now I know that it wasn't done for purpose :)
Will wait for the investigation results.
Thanks again.

Portikus:
[...]

PR #279, opened 2018-07-20
Asynchronous guard conditions and state selectors

rwwilden:
Added a `GuardAsync` to `GuardCondition`, next to the existing `Guard`. This is not an ideal solution yet because it requires two separate implementation paths: `FireAsync` only checks async guards, `Fire` checks synchronous guards (I realize this is a breaking change so I should do some more work to prevent this (done: commit 262c9a1)).

Ideally, guards should always be asynchronous where synchronous guards are wrapped. However, that would be a breaking change I guess because of the divide between sync and async code via `#if TASKS`.

Also included are a couple of `PermitIf` implementations that use async guards. I have not yet implemented the complete async API surface.

Additionally, added support for asynchronous state selectors. This results in an additional field `_destinationAsync` to be added to `DynamicTriggerBehaviour` (similar to `GuardAsync` in `GuardCondition`). When calling one of the `FireAsync` overloads, any synchronous state selector is preferred over the asynchronous state selector to not break existing code.


rwwilden:
Oh, by the way, the work is inspired by what was already happening in the `Support-async-conditions-in-PermitIf` branch by @HenningNT. I just added a few things.

rwwilden:
#189

rwwilden:
#189

rwwilden:
@HenningNT is there something I can do to help get this PR accepted?
I'm currently using the code from my fork on a project that should go into production somewhere in November and this is not an ideal situation :)

oising:
This looks like a good addition. Pinging @HenningNT again :)

HenningNT:
@rwwilden , @oising Sorry about the delay, I haven't had the opportunity to follow up on this.

I'll take a look now ;-)

HenningNT:
@rwwilden Why change OnEntryAsync to OnAsyncEntry? This would break some stuff.

rwwilden:
@HenningNT that's a good question. I suppose because it is not actually an async method. But now it lives beside `OnExitAsync` as well which is confusing and inconsistent.

HenningNT:

PR #293, opened 2018-11-23
Fixes #292 multiple calls of OnEntry when having nested InitialTransitions.

### LINKED ISSUE #292: Nested InitialTransitions calls superstate onEntry twice
Hi,

I use two nested InitialTransitions like in the TestCase `EntersSubStateofSubstate`

I assume the state machine of this test case looks like (BTW the generated dot file looks a lot different):
stateless-test

I modified the test case and added some OnEntry and OnExit actions to keep track of their execution:

```csharp

[Fact]
public void EntersSubStateofSubstate()
{

    sm.Configure(State.A)
        .OnEntry(() => Debug.WriteLine("A OnEntry"))
        .OnExit(() => Debug.WriteLine("A OnExit"))
        .Permit(Trigger.X, State.B);

    sm.Configure(State.B)
        .OnEntry(() => Debug.WriteLine("B OnEntry"))
        .OnExit(() => Debug.WriteLine("B OnExit"))
        .InitialTransition(State.C);

    sm.Configure(State.C)
        .OnEntry(() => Debug.WriteLine("C OnEntry"))
        .OnExit(() => Debug.WriteLine("C OnExit"))
        .InitialTransition(State.D)
        .SubstateOf(State.B);

    sm.Configure(State.D)
        .OnEntry(() => Debug.WriteLine("D OnEntry"))
        .OnExit(() => Debug.WriteLine("D OnExit"))
        .SubstateOf(State.C);

    sm.Fire(Trigger.X);

    Assert.Equal(State.D, sm.State);
}
```

What I expect as output is:
```
A OnExit
B OnEntry
C OnEntry
D OnEntry
```

But what I get is:
```
A OnExit
B OnEntry
C OnEntry
B OnEntry
C OnEntry
D OnEntry
```

In my opinion this are one `B OnEntry` and one `C OnEntry` to much ;)

I browsed a little bit through the code but due to its complexity I didn't find something obvious till now what could cause this behavior.

Any ideas how to fix or work around this?

Best regards and thanks for doing stateless,
ahorn

ahorn42:
Fixes #292 for 'normal' StateMachine. Unfortunately I have no real idea how the 'async' version works, so I only added code in the 'normal' part and I added a test case to test the behavior.

All other tests still pass.

Would be happy if this helps to fix the problem :).

Best regards,
ahorn

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ project.lock.json
 artifacts/
 TestResult.xml
 *.orig
+/.localhistory/*
--- a/src/Stateless/StateMachine.Async.cs
+++ b/src/Stateless/StateMachine.Async.cs
@@ -222,7 +222,12 @@ namespace Stateless
                     {
                         var initialTransition = new Transition(source, newRepresentation.InitialTransitionTarget, trigger);
                         newRepresentation = GetRepresentation(newRepresentation.InitialTransitionTarget);
-                        await newRepresentation.EnterAsync(initialTransition, args);
+
+                        if (!newRepresentation.GetSubstates().Any()) 
+                        {
+                            await newRepresentation.EnterAsync(initialTransition, args);
+                        }
+
                         State = newRepresentation.UnderlyingState;
                     }
                     //Alert all listeners of state transition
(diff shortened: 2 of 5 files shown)


ahorn42:
In the mean we already fixed the async version as well ;)

trickybedna:
Hi @ahorn42 - would you be willing to rebase your PR onto the upstream master so this can be merged? 
I'd be happy to do it if you don't mind adding me as a collaborator on the forked project.

HenningNT:
Thank you for your effort!
Unfourtunately I fixed this without checking if anyone else had fixed it as well.

Issue #290, opened 2018-10-26
Why is class StateGraph (in StateGraph.cs) not public?


frankhommers:
Is there a reason for this? Or can it be changed. I would like to generate the graph myself and feed it to my own diagram component.

mclift:
@frankhommers Closing this issue as I believe it was resolved by #295. Please let me know if it needs more work!

Issue #210, opened 2017-11-03
Visualization API is not extensible


smartcaveman:
Even though Stateless has a built-in diagram generator, I never get to use it.  Instead, I end up having to write a custom implementation to generate DOT syntax for graphviz from the `Stateless.Reflection` objects (or call Stateless internals with reflection).  

Here's why:

1.  `StateGraph` is internal, so the only way to generate the DOT syntax from the current implementation is to call  `UmlDotGraph.Format`.

2.  `UmlDotGraph.Format` does not provide any way to customize the structure or style of the DOT output.  Even though `IGraphStyle` is public, there is no public API that parameterizes the graph style.  A new `UmlDotGraphStyle` instance is created within the method, and there is no way to customize it before it is passed to the `ToGraph` method of the internal `StateGraph`.   

There are multiple ways to fix this.  The simplest and most liberal way would be to make `StateGraph` and its components public so that clients can extend the visualization as they see fit.  A more restrictive option would allow delegates and/or `IGraphStyle` instances to be passed into the `Format` method.  The downside to the second, more restrictive option is that clients would only be able to extend the style but not the structure of the DOT graph.  I would be happy to help implement either option.

smartcaveman:
@HenningNT @nblumhardt

JeremyCade:
@smartcaveman I'll take a look at this over the next, week time permitting.

psibernetic:
Looking at it, i'm not sure why StateGraph became internal. It should be made using only what is in StateMachineInfo, which is intentionally public. There should be an interface on building graphs, like what I originally proposed with ICartographer, so that it can use a strategy pattern, similar to IFormatter.

HenningNT:
How are things going with this one? It seems sensible to do something about it.

smartcaveman:
@HenningNT I did some work on this about a year ago.  If it won't step on anyone's toes I can dig that up and make a PR.  @JeremyCade ?

LoloActemium:
Hi,
+1
I would like to be able to "Customize" DOT Graphs.

frankhommers:
The suggested solution is basically what I'm asking here: https://github.com/dotnet-state-machine/stateless/issues/290

frankhommers:
@smartcaveman Please do so. Make it a pull request!

mclift:
@smartcaveman I think this issue was resolved some time ago. Let me know if this should be reopened.

PR #295, opened 2018-12-22
Make visualization API extensible.

mclift:
This PR is to address Issues #210 and #290 by changing several type declarations from `internal` to `public` so that the graphing classes are more extensible.

Additionally, the `abstract` class `IGraphStyle` has been renamed to `GraphStyleBase` to make it consistent with the [.NET Framework Design Guidelines](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-classes-structs-and-interfaces). Finally a few declaration modifiers have been reordered to make them consistent within the solution (e.g. `override public ...` ->`public override ...`).

Review comments & feedback welcomed!


HenningNT:
I hope it works!

Issue #284, opened 2018-09-09
Unmet conditions are not populated when specifying unhanded trigger action


arublevsky:
Hi, I've found an issue with unmet conditions when specifying unhanded trigger action. Please take a look at the following example:
```
var stateMachine = new StateMachine<State, Trigger>(State.Draft);

stateMachine.Configure(State.Draft).PermitIf(Trigger.GoLive, State.Live, () => false, "Reason");

stateMachine.OnUnhandledTrigger((state, trigger, unmetConditions) =>
{
    if (unmetConditions != null && unmetConditions.Any())
    {
        // This line is not going to be executed
        Console.WriteLine(string.Join(Environment.NewLine, unmetConditions));
    }
 });

stateMachine.Fire(Trigger.GoLive);
```

The `unmetConditions` parameter is always null, so I can't log that data to easily identify the reason of this particular failure. 

Looking at the source code, I found that that's happen because of the condition in StateRepresentation.cs line 59:
```
bool TryFindLocalHandler(TTrigger trigger, object[] args, out TriggerBehaviourResult handlerResult)
{     
    ....

    // Remove those that have unmet guard conditions
    // Guard functions are executed here 
    var actual = possible
    .Select(h => new TriggerBehaviourResult(h, h.UnmetGuardConditions(args))) 
    .Where(g => g.UnmetGuardConditions.Count == 0)                                      <---- here
    .ToArray();

    ....
 }
```

Full example:
[ConsoleApp.zip](https://github.com/dotnet-state-machine/stateless/files/2364205/ConsoleApp.zip)

HenningNT:
Checked, and found not an issue any more.

PR #304, opened 2019-02-13
Brought back guard description when guard prevents transition

rustybox:
Doesn't break NoExceptionWhenPermitIfHasMultipleExclusiveGuardsBothFalse test.

Would appreciate @HenningNT comment that this doesn't cause regression.


dalibor-aleksic-atomia:
@HenningNT Will be this accepted? I would like to have this feature also.
#284 is also helpful in my use case

rustybox:
Will rebase and merge this in tomorrow if there’s no objections

HenningNT:
I'm assuming it works better than before ;-)

Issue #306, opened 2019-03-02
Behaviour is different between FireAsync and Fire When Going from Substate to SuperState


yonkahlon:
Using v4.2.1. The output for TelephoneCallExample is:

> [Status:] OffHook
> [Firing:] CallDialed
> [Status:] Ringing
> [Firing:] CallConnected
> [Timer:] Call started at 2019/03/02 15:47:11
> [Status:] Connected
> Volume set to 2!
> [Status:] Connected
> [Firing:] PlacedOnHold
> [Status:] OnHold
> [Firing:] MuteMicrophone
> Microphone muted!
> [Status:] OnHold
> [Firing:] UnmuteMicrophone
> Microphone muted!
> [Status:] OnHold
> [Firing:] TakenOffHold
> [Status:] Connected
> Volume set to 11!
> [Status:] Connected
> [Firing:] HungUp
> [Timer:] Call ended at 2019/03/02 15:47:11
> [Status:] OffHook
> Press any key...

Change code to use FireAsync and we get:

> [Status:] OffHook
> [Firing:] CallDialed
> [Status:] Ringing
> [Firing:] CallConnected
> [Timer:] Call started at 2019/03/02 15:45:11
> [Status:] Connected
> Volume set to 2!
> [Status:] Connected
> [Firing:] PlacedOnHold
> [Status:] OnHold
> [Firing:] MuteMicrophone
> Microphone muted!
> [Status:] OnHold
> [Firing:] UnmuteMicrophone
> Microphone muted!
> [Status:] OnHold
> [Firing:] TakenOffHold
> [Timer:] Call ended at 2019/03/02 15:45:11    **<-- Here we are calling the super state's [Connected] OnExit**
> [Timer:] Call started at 2019/03/02 15:45:11   **<-- Here we are calling the super state's [Connected] OnEntry**
> [Status:] Connected
> Volume set to 11!
> [Status:] Connected
> [Firing:] HungUp
> [Timer:] Call ended at 2019/03/02 15:45:11
> [Status:] OffHook
> Press any key...

Could the fix for issue [#217](https://github.com/dotnet-state-machine/stateless/issues/217) need to be applied to FireAsync?

HenningNT:
Thank you for the bug fix!

PR #307, opened 2019-03-03
Check if we are transitioning to superstate when using FireAsync

yonkahlon:
Fix for Issue #306  

Replicate logic to Async code so we do not re-transition to superstate when leaving substate


HenningNT:
Looks good, thank you very much for this!

PR #319, opened 2019-05-24
Added support for transition functions

aworx:
Added support for transition functions that gets called after the Exit and before the Entry. It contains a simple test


HenningNT:
Interesting!
Would you mind adding some tests that verify the same behaviour on the Async side as well, please?

aworx:
I don't have much experience in using an async statemachine but I will try my best. Thanks for your feedback

HenningNT:
Come to think of it, I think we shouldn't add this.
We already have exit and entry actions in addition to the OnTrasition event. I don't think we need an additional way of calling a function.

PR #320, opened 2019-06-14
Add additional framework targets

andrewlock:
Thanks for this library! This PR makes minor tweaks to the targeted frameworks to follow best practice advice 🙂

As you've included a .NET Framework 1.x target, best practice advice is to also include a .NET Standard 2.0 target ([as described in this doc](https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/cross-platform-targeting)). This avoids bringing in loads of extra packages on non-NETFX frameworks that support .NET Standard 2.0 (e.g. .NET Core)

Unfortunately, .NET Standard support is all screwed up around `net461`. As you already have a `net45` target, all that's needed to fully fix it is the `net472` target.

From a practical point of view, this shouldn't have any impact on consumers, other than simplifying the package graph on .NET Core!


vip32:
@onuraltun when will this PR be merged? we need netstandard2.0 support

vip32:
can be closed: netstandard2.0 support has already been merged (#334)

nblumhardt:
Updated this to narrow it to .NET 4.7.2 addition only.

WeihanLi:
why add a `net472` target? I think `net45` is enough

Issue #332, opened 2019-10-27
.NetStandard 2.1 support


abuassar:
I have a .netstandar2.1 project, and although stateless package has no dependencies yet it asks for a huge number of packages to be installed:

`
Installing:
Microsoft.NETCore.Platforms.1.1.0

Microsoft.NETCore.Targets.1.1.0

Microsoft.Win32.Primitives.4.3.0

NETStandard.Library.1.6.1

runtime.debian.8-x64.runtime.native.System.Security.Cryptography.OpenSsl.4.3.0

runtime.fedora.23-x64.runtime.native.System.Security.Cryptography.OpenSsl.4.3.0

runtime.fedora.24-x64.runtime.native.System.Security.Cryptography.OpenSsl.4.3.0

runtime.native.System.4.3.0

runtime.native.System.IO.Compression.4.3.0

runtime.native.System.Net.Http.4.3.0

runtime.native.System.Security.Cryptography.Apple.4.3.0

runtime.native.System.Security.Cryptography.OpenSsl.4.3.0

runtime.opensuse.13.2-x64.runtime.native.System.Security.Cryptography.OpenSsl.4.3.0

runtime.opensuse.42.1-x64.runtime.native.System.Security.Cryptography.OpenSsl.4.3.0

runtime.osx.10.10-x64.runtime.native.System.Security.Cryptography.Apple.4.3.0

runtime.osx.10.10-x64.runtime.native.System.Security.Cryptography.OpenSsl.4.3.0

runtime.rhel.7-x64.runtime.native.System.Security.Cryptography.OpenSsl.4.3.0

runtime.ubuntu.14.04-x64.runtime.native.System.Security.Cryptography.OpenSsl.4.3.0

runtime.ubuntu.16.04-x64.runtime.native.System.Security.Cryptography.OpenSsl.4.3.0

runtime.ubuntu.16.10-x64.runtime.native.System.Security.Cryptography.OpenSsl.4.3.0

Stateless.4.2.1

System.AppContext.4.3.0

System.Buffers.4.3.0

System.Collections.4.3.0

System.Collections.Concurrent.4.3.0

System.Console.4.3.0

System.Diagnostics.Debug.4.3.0

System.Diagnostics.DiagnosticSource.4.3.0

System.Diagnostics.Tools.4.3.0

System.Diagnostics.Tracing.4.3.0

System.Globalization.4.3.0

System.Globalization.Calendars.4.3.0

[...]

PR #333, opened 2019-10-27
Update Stateless.csproj to add .netstandard2.1 support

abuassar:
add .netstandard2.1 support

### DIFF
--- a/src/Stateless/Stateless.csproj
+++ b/src/Stateless/Stateless.csproj
@@ -6,7 +6,7 @@
     <NeutralLanguage>en-US</NeutralLanguage>
     <VersionPrefix>4.2.0</VersionPrefix>
     <Authors>Stateless Contributors</Authors>
-    <TargetFrameworks>netstandard1.0;net45;net40</TargetFrameworks>
+    <TargetFrameworks>netstandard2.1;netstandard1.0;net45;net40</TargetFrameworks>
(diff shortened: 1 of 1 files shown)


abuassar:
related issue: 
https://github.com/dotnet-state-machine/stateless/issues/332

Mephistofeles:
.NET Standard 2.1 is supported only by .NET Core 3.0. Why do you want to drop support for older versions, when there is no inherent advantage of using 2.1, when not using any new functions? 
I think it should be changed to .NET Standard 2.0, which is compatible with 2.1 projects.

abuassar:
> 
> 
> .NET Standard 2.1 is supported only by .NET Core 3.0. Why do you want to drop support for older versions, when there is no inherent advantage of using 2.1, when not using any new functions?
> I think it should be changed to .NET Standard 2.0, which is compatible with 2.1 projects.

we can add both netstandard2.0 and netstandard2.1

vip32:
Netstandard 2.0 should suffice. 2.1 not needed as this lib does not use any of its features.

What is holding stateless back regarding the 2.0 targeting? So it can be properly used with netcore 2/3

vip32:
here's a proper netstandard 2.0 only PR #334

vip32:
can be closed: netstandard2.0 support has already been merged (#334)

Issue #288, opened 2018-09-27
TelephoneCallExample: System.ArgumentException : An argument of type System.Int32 is required in position 0


rosdi:
Running the TelephoneCallExample throws out this exception:

```
System.ArgumentException : An argument of type System.Int32 is required in position 0.
   at Stateless.ParameterConversion.Unpack(Object[] args, Type argType, Int32 index)
   at Stateless.ParameterConversion.Unpack[TArg](Object[] args, Int32 index)
   at Stateless.StateMachine`2.TransitionGuard.<>c__DisplayClass4_0`1.<ToPackedGuard>b__0(Object[] args)
   at Stateless.StateMachine`2.TransitionGuard.<>c__DisplayClass17_0.<UnmetGuardConditions>b__0(GuardCondition c)
   at System.Linq.Enumerable.WhereSelectListIterator`2.ToList()
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at Stateless.StateMachine`2.StateRepresentation.<>c__DisplayClass69_0.<GetPermittedTriggers>b__2(TriggerBehaviour a)
   at System.Linq.Enumerable.Any[TSource](IEnumerable`1 source, Func`2 predicate)
   at Stateless.StateMachine`2.StateRepresentation.<>c__DisplayClass69_0.<GetPermittedTriggers>b__0(KeyValuePair`2 t)
   at System.Linq.Enumerable.WhereSelectEnumerableIterator`2.ToArray()
   at Stateless.StateMachine`2.ToString()
   at System.Text.StringBuilder.AppendFormatHelper(IFormatProvider provider, String format, ParamsArray args)
   at System.String.FormatHelper(IFormatProvider provider, String format, ParamsArray args)
   at System.String.Format(String format, Object arg0, Object arg1)
   at StatelessTest.PhoneCall.Print() in C:\Repos\StatelessTest\StatelessTest\PhoneCall.cs:line 116
```

This is very similar (but not identical) to issue https://github.com/dotnet-state-machine/stateless/issues/275

sgoodgrove:
Seems to centre around the call

    .InternalTransition<int>(_setVolumeTrigger, (volume, t) => OnSetVolume(volume))

in PhoneCall.cs

[...]

PR #341, opened 2019-12-05
fix for TelephoneCallExample

fjod:
https://github.com/dotnet-state-machine/stateless/issues/288  

nobody cared to fix it for a year?


HenningNT:
Low on bandwidth :-(

Issue #346, opened 2020-01-16
[Question] Is it possible to access Trigger parameters in `.OnTransitioned()`?


Brondahl:
I think the answer is currently "No". But It wouldbe great to have that confirmed.

Assuming I'm right, then would it be possible to get this added?

smnbackwards:
Hey @Brondahl , I was hoping for the same feature. I took a look at the code and it seemed simple to add, so I made a PR #348 . Not sure it is done in the spirit the maintainers want , but maybe it can temporarily unblock you.

HenningNT:
I'd like to know the use case, why would like to add this feature?

Brondahl:
I don't really see why you'd think that OnTransition() would want to have the Trigger in general, but not the Parameter? If a Trigger is Parameterised then that Param is *part* of that Trigger event?

In my particular case, the answer is Logging.

I have some transitions happening which I would superficially assert weren't possible.

I have a theory that they're caused by parallel calls firing in odd orders.
So I want to attach variables to my Triggers as parameters, at Fire-time and be able to log them passing through OnTransition.

Brondahl:
But to repeat, my particular case is just a single instance. OnTransition is a general method providing a general hook with Trigger information. Therefore technical and resourcing limitations aside, it feels appropriate that it should have *all* the Trigger information available to it, including the Parameter info.

smnbackwards:
My use case is very similar to @Brondahl and I agree with his justification for adding parameters to OnTransition.

I'm using OnTransition to allow me to have a single log line that gives me all the information about how the state is being changed. It's super useful for reproducing bugs etc since I know the starting state and every transition, so I can setup the initial state and replay all the triggers. However, since I have a number of parameterized triggers it isn't very useful without the parameters. I believe OnTransitioned isn't really representing the transition that happened faithfully if the parameters aren't included, and I'd also argue that the Transition class isn't being faithful either if the parameters aren't accessible.

HenningNT:
I agree that the parameters are very handy to have during debugging, for logging purposes.
I'm vary of adding new features that are not strictly necessary, but this one seems to be useful.

HenningNT:
If we get unit tests for the PR, then I'd be happy to merge it in.

smnbackwards:
Sorry for the delay. I'm on it and haven't forgotten. I'll get a PR out in the next day or two.

smnbackwards:
Updated PR is out. Let me know if you think there are missing test cases.

[...]

PR #348, opened 2020-01-24
A basic attempt at adding Trigger parameters to the OnTransitioned

smnbackwards:
This was a very quick, basic attempt to add Trigger parameters to the OnTransitioned
event.

This was requested in  #346 and something I was interested in as well.
This change is more of a proof of concept to see if the maintainers consider
this type of an approach acceptable. If so, I would add tests and ensure that
if the parameters are not specified they are stored as an empty array on the
Transition object.


HenningNT:
Looks good!
Would you mind adding unit tests that cover this code?

Will this be a drop in replacement, or will the users of stateless have to change how they use it?

Issue #177, opened 2017-06-06
Should Activation/deactivation be executed on every state exit/entry?


HenningNT:
I thought the purpose of the Activation/deactivation was to be run if needed, not on every state change. For instance, maybe when an IoT speaker has its power restored it would start up in the Playing state, but the connection is not there anymore. The OnActive method could then re-connect to the stream it was playing before it lost power.

As it is implemented at the moment it duplicates the entry/exit actions, and I suspect not very useful. I would like to know you think about it.

nblumhardt:
Hi Henning - I think this is by design, if the `OnActivate()` handler isn't called when transitioning into a state, every `OnActivate()` entry ends up being duplicated as an `OnEntry()` handler. The main difference is the additional `OnActivate()` call triggered by `StateMachine.Activate()`.

I haven't used it in anger however, so I may be overlooking the obvious. @yevhen knows this feature best so might have some input here.

HenningNT:
I hope he has some input, as I'm pretty sure the concept involved activation/deactivation when needed (called externally), not on every entry/exit.

yevhen:
Activation is different from state change. State machine can be resurrected (materialized) in some finite state (i.e. initial state) and no OnEntry handlers will be executed, but you still want to be notified that it is now *active*, so you can start a timer for example. The current implementation mirrors UML 2.0 Active State definition (two different handlers for state transition and state activation).

@HenningNT the use-case you are describing might be better modeled with state transitions like `Connected/Disconnected/Playing` and then you can use `OnActivate` handler to play the stream.

HenningNT:
The example is a bit contrived, yes.

But should a state activation/deactivate function be called on every state transition, or just when needed?

yevhen:
Yes, since it is an activation of a new state.

HenningNT:
But... we have entry actions for that?

yevhen:
See https://github.com/dotnet-state-machine/stateless/issues/177#issuecomment-308240430

> State machine can be resurrected (materialized) in some finite state (i.e. initial state) and no OnEntry handlers will be executed, but you still want to be notified that it is now active, so you can start a timer for example.

HenningNT:
Yes, so activation isn't the same as transitioning to a new state, and executing the Entry action. They need to be two different things.

Consider a larger example:
[...]

PR #356, opened 2020-02-21
Activate and deactive only when requested

HenningNT:
Breaks old behaviour.
Bump version to v5.0.


HenningNT:
Appveyor build fails due to branch name containing "#177" ...

Issue #305, opened 2019-02-18
DynamicTransitionInfo does not contains all possible states when using PermitDynamicIf configuration


audricpod:
Hi there,

I have been using for a few weeks Stateless and it is working great !
But the default `UmlDotGraph.Format` method is not enough for me and I'm a creating a new one (for isolating sub graphe) using the Reflection API.

And I am facing a problem for generating my statemachine graph, when using `PermitDynamicIf `state configuration, the `PossibleDestinationStates` property is not accessible in the `DynamicTransitionInfo` and also is null

Is it a bug ? or volontary restriction ?
Is there another way to get all the available dynamic state ?

thanks in advance for helping

Audric

audricpod:
With a sample it is always better

Here is my state configuration :

```
           Configure(StateAA)
                .SubstateOf(StateA)
                .OnExit(() => anything())
            .PermitDynamicIf(Triggers.ValidateState,
                               () => IsFInished ? StateFinished : StateX ,
                               () => State.IsStateValid);
```


and here my graph function :

        public string GetSubgraph(string superStateStr)
        {
            States? SuperState = Enum.Parse(typeof(States), superStateStr) as States?;			
            if (!SuperState.HasValue)
            {
                return "";
            }

            StringBuilder graphSb = new StringBuilder();
            graphSb.AppendLine(@"digraph G {");
            graphSb.AppendLine();
			
			[...]


			var allSubstates = GetInfo().States.Where(x => x.Superstate != null &&
                                                      (x.Superstate.UnderlyingState is MyState myState) &&
                                                      myState.CurrentState == SuperState);
            // Transitions
            foreach (var s in allSubstates)
            {
                var myState = (s.UnderlyingState as MyState);

                foreach (var t in s.Transitions)
                {
                    if (t is FixedTransitionInfo fixedTr)
                    {
                        graphSb.AppendLine($"{myState.CurrentState} -> {fixedTr.DestinationState};");
                    }
                    else if (t is DynamicTransitionInfo dynamicTr)
                    {
                        // KO - github issue 
                        // https://github.com/dotnet-state-machine/stateless/issues/305
                        //foreach (var possibleState in dynamicTr.PossibleDestinationStates)
                        //{
                        //    graphSb.AppendLine($"{myState.CurrentState} -> {possibleState.ToString()};");
                        //}

                    }
                }                
            }
            graphSb.Append("}");

            return graphSb.ToString();
        }

mclift:
[...]

Issue #267, opened 2018-06-01
Exception from OnEntry action should not result in queued trigger being executed after next trigger


Snaaio:
I would like to design a state machine in which an exception thrown from an OnEntry action can propagate to the caller firing the trigger. Unfortunately I saw some weird behavior regarding the queuing of triggers.

As a workaround I upgraded to Stateless v4.2.0 and configured my statemachine to use FiringMode.Immediate, but it would be nice to see the default FiringMode.Queued play a bit nicer with exceptions being thrown.

Possible options to better handle this:
1. Catch exception and execute queued triggers before rethrowing
1. Catch exception and clear queued triggers before rethrowing
1. Do not catch exception and retain queued triggers (current behavior)
   Add new method(s) on StateMachine to:
   A. Fire queued triggers
   B. Clear queued triggers

```
enum State { Idle, Initializing, Ready, Error }
enum Trigger { Init, Ready, Error }

[Test]
public void NUnitTestStateless_FireErrorTriggerFromOnEntryActionAndThenThrow_ShouldNotExecuteQueuedErrorTriggerAfterNextTrigger()
{
    var stm = new StateMachine<State, Trigger>(State.Idle);
    stm.Configure(State.Idle)
        .Permit(Trigger.Init, State.Initializing);
    stm.Configure(State.Initializing)
        .Permit(Trigger.Ready, State.Ready)
        .Permit(Trigger.Error, State.Error)
        .OnEntry(() =>
        {
            try
            {
                throw new InvalidOperationException("Test exception.");
            }
            catch
            {
                stm.Fire(Trigger.Error);
                throw;
            }
        });
    try
    {
        stm.Fire(Trigger.Init);
    }
    catch
    {
        // Normally the exception would be handled here
    }

    stm.Fire(Trigger.Ready); // Fails with System.InvalidOperationException : No valid leaving transitions are permitted from state 'Ready' for trigger 'Error'. Consider ignoring the trigger.

    Assert.That(stm.State, Is.EqualTo(State.Ready));
}
```

Snaaio:
@HenningNT The weird behaviour was:
1. Initially a caller fires a first trigger A (from externally), which results in a state transition and execution of the corresponding OnEntry action
1. From that OnEntry action a trigger is fired, which is queued internally by stateless
1. Later in that same OnEntry action, an exception is thrown which is unhandled inside the OnEntry action and propagates through stateless to the caller of the initial trigger
[...]

PR #362, opened 2020-03-26
Release 5.1.0

HenningNT:
Added
- Added missing possible destination states to all PermitDynamic and PermitDynamicIf, see issue #305.

Fixed
- Fixed trigger execution order issue if there are uncaught exceptions. Ref issue #267.
- Fixed issue #272  and #275 . Triggers with no parameters threw exceptions when checking if they could be fired, or retrieved with GetPermittedTriggers.


HenningNT:
I'd like to release a new version. I've fixed a few bugs and added some optional parameters.
(I don't understand why committs all the way back to 2018 are included in the PR)

If there are no objections I'll merge it in a couple of days.

PR #369, opened 2020-04-14
Even more fluent syntax

HenningNT:
Added a more flexible, generic fluent syntax.
Also added an optional Action to the transitions.

I took inspiration from PR #173, and added what I thought would work.


HenningNT:
@nblumhardt I'd really like your opinion about this one :-)

src/Stateless/DestinationConfiguration.cs
```diff
@@ -0,0 +1,93 @@
+﻿using System;
+
+namespace Stateless
+{
+    partial class StateMachine<TState, TTrigger>
+    {
+        /// <summary>
+        /// This class containd the required trigger information for a transition.
```
mclift, line 8: Typo in "containd"

src/Stateless/TransitionConfiguration.cs
```diff
@@ -0,0 +1,121 @@
...
+                return new DestinationConfiguration(this, ttb, _representation);
+            }
+
+            /// <summary>
+            /// Creates a new internal transition. This transition, when triggered, not cause any state change,
+            /// nor will the Exit or Entry actions of the state be executed. Only the action configured will be
+            /// executed.
+            /// </summary>
+            /// <returns></returns>
+            public DestinationConfiguration Internal()
```
mclift, line 59: There are times this would have been really useful on a previous project! However, is there some way that internal transitions could be represented differently in the dot graph to show that the entry & exit functions won't be executed?

src/Stateless/DestinationConfiguration.cs
HenningNT: fixed

src/Stateless/TransitionConfiguration.cs
HenningNT: I guess so, someone would have to add it... In QM it looks like this:
image

HenningNT: This is QM, BTW:
https://www.state-machine.com/qm/

I've used QP for microcontroller state machines, and designed the state machines in QM.

nblumhardt:
Some nice features in the new syntax, @HenningNT 👍 ... the main downside would be the tradeoff between accumulating redundant code/inconsistent API naming ("permit" a trigger vs. "transition" on a trigger), and breaking changes. Unsure which I'd choose, there's some merit to keeping things moving forward and trimmed-down 🤔 

I'm still low on time to contribute here, so apologies for not digging in deeper - good luck with it!

HenningNT:
I considered using Permit, but I think it's better to change the name to Transition since it's a new syntax with different functionality. This way the confusion should be reduced somewhat.

I would like to remove the "old" syntax, and only use the "new". There would be a transition period where the old and new syntax would haver to co-exist.
After the new syntax has been fully deployed and tested, the old could be labeled deprecated, and after a time removed.

If the team @dotnet-state-machine/reviewers is OK with it?

ffMathy:
I think this is awesome! @HenningNT is this completed? Because if so, I can merge it for you. And if not, I can finish it for you.

HenningNT:
I think it is complete, all the tests pass...

But it is a completely new syntax, so it would be a new major version number.  I was unsure what to do with it...

leeoades:
Fyi - I developed a different fluent interface that I found easier to understand as a set of extention methods and builder classes. Under the hood, they call into the unchanged stateless interface.

My plan is to release it as its own package, basically making it optional. It also allows me to document the different interface without muddying the water here.

Perhaps the same approach could be used here?
If at some point everyone is using one of the different fluent interfaces then it could be considered then making it the default?
Alternatively, keeping them separate offers consumers choice.

Issue #372, opened 2020-04-24
_onTransitionedEvent Is fired in a different order for FireAsync


juliusl:
I noticed that _onTransitionEvent was called in reverse when driving parts of the state machine with FireAsync. 

Looking at the source I noticed that the ordering is different for FireAsync vs Fire. 

In [Fire](https://github.com/dotnet-state-machine/stateless/blob/db8a0d9577f71bde8918c4c656f4362f5bfa273e/src/Stateless/StateMachine.cs#L427)
```
          //Alert all listeners of state transition
            _onTransitionedEvent.Invoke(transition);
            var representation = EnterState(newRepresentation, transition, args);
```

In [FireAsync ](https://github.com/dotnet-state-machine/stateless/blob/db8a0d9577f71bde8918c4c656f4362f5bfa273e/src/Stateless/StateMachine.Async.cs#L231)
```
                var newRepresentation = GetRepresentation(transition.Destination);
                await newRepresentation.EnterAsync(transition, args);

// omitted
                // Alert all listeners of state transition
               await _onTransitionedEvent.InvokeAsync(new Transition(source, destination, trigger, args));
```
Currently written, it will only call _onTransitionedEvent after OnEnter methods are called. If the subsequent OnEnterAsync are called that call FireAsync for a state transition, that means that _onTransitionedEvent won't be called until the transition finishes. 

For **FiringMode.Queued** this is not an issue since ordering is preserved, but for **FiringMode.Immediate** it will result in incorrect ordered onTransitionedEvent calls even though the transition order is correct.

juliusl:
PR for your review, https://github.com/dotnet-state-machine/stateless/pull/373

HenningNT:
Thanks a lot!
I'll update the changelog, and get a new release out.

juliusl:
Thanks for the merge!

PR #373, opened 2020-04-24
Fix _onTransitionedEvent.InvokeAsync ordering

juliusl:
Address #372


HenningNT:
Awesome, thanks for the fix!

Issue #380, opened 2020-06-22
PermitDynamicIf call ambiguous if trigger parameter is a state


DeepakParamkusam:
Compiler complains that PermitDynamicIf calls in below code are ambiguous-
```
static void Main(string[] args)
{
    var fsm = new StateMachine<State, Trigger>(State.Off);
    StateMachine<State, Trigger>.TriggerWithParameters<State> pressTrigger = fsm.SetTriggerParameters<State>(Trigger.Press);
    StateMachine<State, Trigger>.TriggerWithParameters<State> smashTrigger = fsm.SetTriggerParameters<State>(Trigger.Smash);

    fsm.Configure(State.Off)
        .PermitDynamicIf(pressTrigger, state => { return state; })  // CALL AMBIGUOUS
        .PermitDynamicIf(smashTrigger, GetState);                   // CALL AMBIGUOUS

    while (true)
    {
        Console.WriteLine("Switch is in state: " + fsm.State);
        var pressed = Console.ReadKey(true).KeyChar;

        if (pressed != 'a') break;

        fsm.Fire(smashTrigger, State.Destroyed);
    }
}

public static State GetState(State state)
{
    return state == State.Off ? State.Destroyed : State.On;
}

public enum State
{
    On,
    Off,
    Destroyed
}

public enum Trigger
{
    Press,
    Smash
}
```

Compiler error-

`The call is ambiguous between the following methods or properties: 'StateMachine<TState, TTrigger>.StateConfiguration.PermitDynamicIf<TArg0>(StateMachine<TState, TTrigger>.TriggerWithParameters<TArg0>, Func<TArg0, TState>, DynamicStateInfos, params Tuple<Func<bool>, string>[])' and 'StateMachine<TState, TTrigger>.StateConfiguration.PermitDynamicIf<TArg0>(StateMachine<TState, TTrigger>.TriggerWithParameters<TArg0>, Func<TArg0, TState>, DynamicStateInfos, params Tuple<Func<TArg0, bool>, string>[])'`

The ambiguity seems to be in the optional params. Am I doing something wrong? 

Using .NET Core 3.1, Stateless v5.1.2

DeepakParamkusam:
It compiles if the tuple is explicitly cast -
```
fsm.Configure(State.Off)
    .PermitDynamicIf(pressTrigger, state => state, null, (Tuple<Func<bool>,string>)null)
[...]

PR #384, opened 2020-08-11
Dev

scottgorlick:
issue #380 resolution:
 - add Method signature with no optional parameters
 - new method calls into specified overload in issue comments using explicit types builds empty Tuple<Func<bool>, string>[] to satisfy params Tuple<Func<bool>, string>[] defaults


HenningNT:
Awesome, thank you for making the world a tiny bit better!

Issue #394, opened 2020-10-05
Add support for an OnTransitioned event which happens after OnEntry etc have been invoked


smnbackwards:
I ran into a scenario when writing a Blazor app where I wanted to update clients whenever a transition (including re-entry) occurred.
(And similarly to this old Issue I worked on #346, I believe it's a good entry point for setting up logging to reproduce bugs)

I noticed that OnTransitioned seems to purposely be being called before the state mutator is invoked - there is a test called "TheOnTransitionEventFiresBeforeTheOnEntryEvent". In my scenario this meant that my clients were updated, but nothing had changed so they were always 1 step behind. 

I would like to propose two things:
1. Short term, a OnTransitionCompleted event handler (I have a draft PR for this to show the general idea - #395)
2. Long term - make a breaking change to rename OnTransitioned to OnTransitioning and OnTransitionCompleted to OnTransitioned 

I believe others would also benefit from 1 and am happy to make the change for that now.

I think 2 makes sense, but might not be a high priority right now. It is confusing to me for OnTransitioned to be called before the state has actually been changed or the OnEntry events have been invoced. OnTransitioning implies the transition is about to happen but hasn't finished yet.

HenningNT:
It's a good point! Naming should be consistent with behaviour.
To start with we should not break things when adding new things, but I am partial to the idea of renaming the event in the future.

mnuess:
Today I ran into the same situation. And I totally expected the OnTransitioned to be called after the OnEntry, simply due to the naming. 
I agree to both of your points, and I'm hoping for an "Completed"-Event soon.

smnbackwards:
@mnuess I have been using a local copy of stateless with my draft PR changes from last week and haven't had any issues. Feel free to give that a try if you are ok with building it yourself. 
@HenningNT, I marked #395 as ready for review. It has the async code path + tests for the short term solution of adding OnTransitionCompleted

DeepakParamkusam:
Observed this issue in our code as well. Hopefully @smnbackwards's PR gets merged soon.

mnuess:
@smnbackwards Thank you for your work. I'll give it a try. But I'm still hopping for a merge quite soonish.

HenningNT:
Will be released in 5.1.6 as soon as we can update the nuget API key ;-)

PR #395, opened 2020-10-05
Add OnTransitionCompleted event which fires after OnTransitioned

smnbackwards:
This is a quick proof of concept for implementing a OnTransitioned event
which is fired after the OnEntryFrom actions have been invoked.

Tests + Async path still todo - I want to check that the maintainer thinks this is worthwhile first
I have described my reasoning for this PR in #394


smnbackwards:
I've added the async code path as well as a number of tests.
Let me know if you think there is a scenario missing coverage.

HenningNT:
I think we need this!

Issue #398, opened 2020-10-16
Unmet guard collection not set by OnUnhandledTrigger if transition guard of substate fails


DeepakParamkusam:
```
static void Main(string[] args)
{
    var fsm = new StateMachine<State, Trigger>(State.B);
    fsm.OnUnhandledTrigger((s, t, c) => Console.WriteLine(c?.First() ?? "UnmetGuards = null"));

    fsm.Configure(State.B)
        .SubstateOf(State.A)
        .PermitIf(Trigger.X, State.C, () => false, "B->C guard failed");

    fsm.Fire(Trigger.X);

    Console.ReadKey(true);
}

public enum State
{
    A,
    B,
    C
}

public enum Trigger
{
    X
}
```
**Expected output** = "B->C guard failed"
**Actual output** = "UnmetGuards = null"

Using Stateless v5.1.3 and .NET Core 3.1

HenningNT:
Thanks a lot!
I've merged this and will try to get a release out.

DeepakParamkusam:
Fixed in v5.1.4

PR #399, opened 2020-10-16
Fix for unmet guard collection not being set by OnUnhandledTrigger 

DeepakParamkusam:
As described in #398, unmet guard description collection is not set by OnUnhandledTrigger if substate transition guard fails.  

This happens because _handler_ parameter in _TryFindHandler_ method is always set to super state handler.


DeepakParamkusam:
I'm uncertain whether a test is required for _OnUnhandledTriggerAsync_ as well. I can add one if required.

HenningTorsteinsenBouvet:
Very nice!

PR #403, opened 2020-11-15
Fix/cls compliance and eol

HenningNT:
Set Assembly CLS compliant.
Set target framework to netcore2.1 instead of 2.2, since 2.2 is out of service.
Also fixed some suggestions by the editor and analyzers, and fixed a bit of spelling.


HenningNT:
Tests fail because I used a / in the branch name...

PR #405, opened 2020-11-21
Release 5.1.6

HenningNT:


HenningNT:
@nblumhardt I think the AppVeyor API key has expired, would you mind looking into it?

Issue #413, opened 2021-01-05
Initial transition to substate does not trigger OnTransitioned event


DeepakParamkusam:
Consider:
```
public static void Main(string[] args)
{
    var sm = new StateMachine<State, Trigger>(State.A);
    sm.OnTransitioned(t => Console.WriteLine($"OnTransitioned = {t.Source}->{t.Destination}"));

    sm.Configure(State.A)
       .Permit(Trigger.X, State.B);

    sm.Configure(State.B)
        .InitialTransition(State.C);

    sm.Configure(State.C)
        .SubstateOf(State.B);

    sm.Fire(Trigger.X);
    
    Console.WriteLine($"Final state = {sm.State}");
}

public enum State { A, B, C }

public enum Trigger { X }
```

Output: 
```
OnTransitioned = A->B
Final state = C
```

Is this intended behavior? In my opinion, initial transition to state C should also call OnTransitioned handler and the output should be:
```
OnTransitioned = A->B
OnTransitioned = B->C
Final state = C
```

Using Stateless 5.1.5 in .NET Core 3.1

HenningNT:
I think it should be...
```
OnTransitioned = A->C
Final state = C
```

This is a bug.

DeepakParamkusam:
OnTransitionCompleted event is not available on v5.1.5 (latest nuget). However, I pulled the dev branch and ran the same test using OnTransitionCompleted event and the output is same:
```
OnTransitionCompleted = A->B
Final state = C
```

HenningNT:
I see, looks like I have a bit more to do :-)

HenningNT:
Fixed in 5.1.7

PR #414, opened 2021-01-05
Added OnTransitioned event for initial transition

DeepakParamkusam:
Initial fix for issue #413. Here I assumed that the correct output is:
```
OnTransitioned = A->B
OnTransitioned = B->C
Final state = C
```

### DIFF
--- a/src/Stateless/StateMachine.Async.cs
+++ b/src/Stateless/StateMachine.Async.cs
@@ -229,6 +229,10 @@ namespace Stateless
                     {
                         var initialTransition = new InitialTransition(source, newRepresentation.InitialTransitionTarget, trigger, args);
                         newRepresentation = GetRepresentation(newRepresentation.InitialTransitionTarget);
+
+                        // Alert all listeners of initial state transition
+                        await _onTransitionedEvent.InvokeAsync(transition);
+
                         await newRepresentation.EnterAsync(initialTransition, args);
                         State = newRepresentation.UnderlyingState;
                     }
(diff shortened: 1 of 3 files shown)


DeepakParamkusam:
@HenningNT, if you can confirm that the correct output is `OnTransitioned = A->C`, I'll make the required changes.

`A->C` makes sense to me; as there is only one trigger, so there should be one transition.

DeepakParamkusam:
After thinking about it a bit, I believe `A->C` event should be raised by `OnTransitionCompleted` and `A->B` and `B->C` events by `OnTransitioned`. This would also not break any existing code.

Test code:
```
public static void Main(string[] args)
{
    sm.OnTransitioned(t => Console.WriteLine($"OnTransitioned = {t.Source}->{t.Destination}"));
    sm.OnTransitionCompleted(t => Console.WriteLine($"OnTransitionCompleted = {t.Source}->{t.Destination}"));

    sm.Configure(State.A)
       .Permit(Trigger.X, State.B);

    sm.Configure(State.B)
        .InitialTransition(State.C);

    sm.Configure(State.C)
        .SubstateOf(State.B);

    sm.Fire(Trigger.X);
    
    Console.WriteLine($"Final state = {sm.State}");
}

public enum State { A, B, C }

public enum Trigger { X }
```

Output:
```
OnTransitioned = A->B
OnTransitioned = B->C
OnTransitionCompleted = A->C
Final state = C
```

HenningNT:
Thanks alot, I appreciate your contribution!

Issue #422, opened 2021-01-28
IgnoreIf Guard description is raised when PermitReentryIf Guard Fails


brcaswell:
I'm getting the incorrect\unexpected `IgnoreIf` guard description when testing a `PermitReentryIf` guard failing (on a chess related program I'm in the initial stages on).

Here are brief snippets of both the state-machine configuration and the calling of methods on the type, which fire off the trigger with parameter.

###  State-Machine Configuration of `Square` for handling member `ChessPiece` instance changing.

```c#

_machine.Configure(SquareState.Occupied)
	.IgnoreIf(_occupyTigger, (piece) => piece.Identifier.Equals(ChessPiece.Identifier), "Same piece occupying square")
	.PermitReentryIf(_occupyTigger, (piece) => !piece.BelongsTo.Equals(ChessPiece.BelongsTo), "Cannot occupy square with another piece owned by the same player!")
	.OnEntryFrom(_occupyTigger, SetChessPiece).Permit(SquareStateTrigger.Unoccupy, SquareState.Unoccupied);

 ```
 
 ## Program Main()
 ```c# 

square.Occupy(firstWhitePiece);
square.Occupy(firstWhitePiece);
square.Occupy(secondWhitePiece); //guard related exception expected (but guard description is not correct)
square.Occupy(firstBlackPiece);

 ```
 
 I put together a dotnetfiddle to reproduce the issue.
 
## [Reproduction Link - dotnetfiddle][1]

  [1]: https://dotnetfiddle.net/linW5Y

**Comment by DeepakParamkusam** [comment]
Shorter code to reproduce the issue (on .NET Core 3.1, Stateless 5.1.8) -
```
static class Program
{
    private static void Main(string[] args)
    {
        var fsm = new StateMachine<State, Trigger>(State.A);
        fsm.OnUnhandledTrigger((_, t, c) => Console.WriteLine($"{string.Join(",", c ?? Enumerable.Empty<string>())} on trigger {t}"));
        fsm.OnTransitionCompleted(t => Console.WriteLine($"{t.Source} -> {t.Destination} on {t.Trigger}"));

        fsm.Configure(State.A)
            .Permit(Trigger.X, State.B);

        fsm.Configure(State.B)
            .IgnoreIf(Trigger.X, () => false, "IgnoreIf guard failed")
            .PermitReentryIf(Trigger.X, () => false, "PermitReentryTrigger guard failed");   

        fsm.Fire(Trigger.X);
[...]

PR #423, opened 2021-01-29
Add all guard descriptions when multiple guards fail for same trigger

DeepakParamkusam:
As described in #422 , when multiple guards fail for same trigger, only the first guard description is returned. In this PR, I added all the other guard descriptions to the returned result.


HenningNT:
Thanks!
I'll get this out in the 5.1.9 when we get the nuget api key sorted.

PR #441, opened 2021-03-12
Fix StateInfo.Transitions NullReferenceException for states without transitions

### LINKED ISSUE #440: StateInfo.Transitions generates exception if there are no Transistions
It's this line of code:
https://github.com/dotnet-state-machine/stateless/blob/dev/src/Stateless/Reflection/StateInfo.cs#L146

That line should IMO return an empty set if any of the child sets are empty (or null).

You get this exception for a state without any possible Transitions:
```
System.ArgumentNullException: Value cannot be null. (Parameter 'first')
   at System.Linq.ThrowHelper.ThrowArgumentNullException(ExceptionArgument argument)
   at System.Linq.Enumerable.Concat[TSource](IEnumerable`1 first, IEnumerable`1 second)
   at Stateless.Reflection.StateInfo.get_Transitions()
(...)
```

That state is defined as for example:
```
 _machine.Configure(State.Error)
```

So it's a final state.

One possible solution is to add some null check in the getter of have the most basic constructor always put a `Enumerable.Empty<Type>();` in every sub `IEnumerable` in the `StateInfo` class.

EraYaN:
This returns `Enumerable.Empty<TransitionInfo>()` if both are `null` and still works if either of the `IEnumerable`'s are `null`. When both are not `null` behaviour is unchanged.

Closes #440

### DIFF
--- a/src/Stateless/Reflection/StateInfo.cs
+++ b/src/Stateless/Reflection/StateInfo.cs
@@ -143,7 +143,7 @@ namespace Stateless.Reflection
         /// <summary> 
         /// Transitions defined for this state.
         /// </summary>
-        public IEnumerable<TransitionInfo> Transitions { get { return FixedTransitions.Concat<TransitionInfo>(DynamicTransitions); } }
+        public IEnumerable<TransitionInfo> Transitions { get { return (FixedTransitions ?? Enumerable.Empty<FixedTransitionInfo>()).Concat<TransitionInfo>(DynamicTransitions ?? Enumerable.Empty<DynamicTransitionInfo>()); } }
 
         /// <summary>
         /// Transitions defined for this state.


HenningNT:
Hi, do you have a unit test that demonstrates this, I'm unable to reproduce it?

EraYaN:
I have to admit that I can't really figure out anymore how I broke it. I'll give it a couple more tries. I think at some point the StateInfo objects didn't get the AddRelationships functions called, so there were nulls for certain states and that broken trying to serialize to json.

HenningNT:
OK, so if we can't reproduce it then I'd rather not merge it. Don't fix it if it's working, you know...

EraYaN:
It's been too long and I can't recall how we broke it, and the change was before the initial commit. So I'll close this PR and the issue.

PR #443, opened 2021-04-06
Added CanFire overload to return unmet guard descriptions

DeepakParamkusam:
Currently, unmet guard descriptions are accessible only via `OnUnhandledTrigger` action override _after_ the trigger is fired. I've come across multiple cases where it would be useful to check if any guard is unmet _before_ firing the trigger.  I've added an overload to `CanFire` method to get the unmet descriptions in case a guard fails.


HenningNT:
Well done, thank you for taking the time!

Issue #444, opened 2021-04-07
[Bug+fix] Inconsistency in sync/async statemachine execution


vrenken:
Hi,

during the easter holidays the wetter was bad which gave me some time to work on a [PlantUML 2 Stateless Roslyn code generator](https://github.com/vrenken/EtAlii.Generators/blob/main/Source/EtAlii.Generators.Stateless.Tests/README.md).

It's amazing to see how such a great library as Stateless benefits when being configured in a more visually way - It completely speeds up the whole development process as the mental model is far easier to grasp. 

However, during this testing I also wanted to make sure that both synchronous and asynchronous state machines work as expected, and in doing so I think I might have spotted an issue in how asynchronous calls in Stateless work with superstates/substates. 

It's best to demonstrate with two 'identical' state machines.

First the synchronous one (which works as expected):
```cs
  [Fact]
  public void SuperStateShouldNotExitOnSubStateTransition_WhenUsingSyncTriggers()
  {
      // Arrange.
      var sm = new StateMachine<State, Trigger>(State.A);
      var record = new List<string>();

      sm.Configure(State.A)
          .OnEntry(() => record.Add("Entered state A"))
          .OnExit(() => record.Add("Exited state A"))
          .Permit(Trigger.X, State.B);
      
      sm.Configure(State.B) // Our super state.
          .InitialTransition(State.C)
          .OnEntry(() => record.Add("Entered super state B"))
          .OnExit(() => record.Add("Exited super state B"));

      sm.Configure(State.C) // Our first sub state.
          .OnEntry(() => record.Add("Entered sub state C"))
          .OnExit(() => record.Add("Exited sub state C"))
          .Permit(Trigger.Y, State.D)
          .SubstateOf(State.B);
      sm.Configure(State.D) // Our second sub state.
          .OnEntry(() => record.Add("Entered sub state D"))
          .OnExit(() => record.Add("Exited sub state D"))
          .SubstateOf(State.B);

      
      // Act.
      sm.Fire(Trigger.X);
      sm.Fire(Trigger.Y);
      
      // Assert.
      Assert.Equal("Exited state A", record[0]);
      Assert.Equal("Entered super state B", record[1]);
      Assert.Equal("Entered sub state C", record[2]);
      Assert.Equal("Exited sub state C", record[3]);
      Assert.Equal("Entered sub state D", record[4]);
  }
```

And the asynchronous one (which doesn't work as expected)
```cs
  [Fact]
[...]

PR #445, opened 2021-04-07
Fix for Inconsistency in sync/async statemachine execution

vrenken:
This commit fixes an inconsistency between the StateRepresentation.Async.cs and StateRepresentation.cs, in that the first did not have the same checks as the latter.  (see #444 for more details)  

There are two unit tests to prove that this fix improves Stateless. They represent the same state machine, both as an asynchronous and synchronous variation. Without the fix the two state machines give different results.


HenningNT:
Very nice, thank you for your contribution!

PR #447, opened 2021-04-19
Added support for spaces in state/trigger names in Graphviz node graphs by wrapping them in escaped quotes

drdavient:


HenningNT:
Thank you for this addition!
It appears that some unit tests fail at the moment, when those are fixed I'd me very happy to merge in this!

drdavient:
Thanks - I've fixed the unit tests and added another one to test the escaped spaces :)

HenningNT:
Awesome work!

PR #452, opened 2021-05-22
Make code async-first 

sunghwan2789:
This PR deletes many duplicate codes that support both synchronous and asynchronous methods. It is done by introducing `EventCallback`, part of ASP.NET Core, prioritizing asynchronous code, and making synchronous code proxy it.

The interesting thing is that asynchronous code runs synchronously before it executes Task.Run or async I/O code.
As a result, the whole methods are acting the same as before.

### DIFF
--- a/src/Stateless/ActivateActionBehaviour.cs
+++ b/src/Stateless/ActivateActionBehaviour.cs
@@ -24,7 +24,7 @@ namespace Stateless
 
             internal Reflection.InvocationInfo Description { get; }
 
-            public virtual Task Execute()
+            public virtual Task ExecuteAsync()
             {
                 return _callback.InvokeAsync();
             }
--- a/src/Stateless/ActivateActionBehaviour.cs
+++ b/src/Stateless/ActivateActionBehaviour.cs
@@ -5,9 +5,16 @@ namespace Stateless
 {
     public partial class StateMachine<TState, TTrigger>
     {
-        internal abstract class ActivateActionBehaviour
+        internal class ActivateActionBehaviour
         {
             readonly TState _state;
+            private readonly EventCallback _callback;
+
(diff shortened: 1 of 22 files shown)


src/Stateless/ActivateActionBehaviour.cs
```diff
@@ -5,64 +5,34 @@ namespace Stateless
 {
     public partial class StateMachine<TState, TTrigger>
     {
-        internal abstract class ActivateActionBehaviour
+        internal class ActivateActionBehaviour
         {
             readonly TState _state;
+            private readonly EventCallback _callback;
```
anatoly-kryzhanovsky, line 11: use EventCallback is not enought good solution. library can used in simple console host application that does not have this dependency

HenningNT:
Firstly, I must say I appreciate your hard work!

I'll take a look at this when I have time, as it changes quite a lot of files.

leeoades:
I know this was done a while ago but I like the idea of it a lot.

(I checked out the PR and obviously, the master branch has moved on since then so I wasn't trying to rebase it.)

I agree with the sentiment completely. It would remove all of the sync and async duplication.

My initial thoughts were that instead of needing EvenCallback, we could simply wrap all sync actions like so:
(plus the argument versions obviously)
```
public static class ActionExtensions
{
    public static Task ToAsync(this Action action)
    {
        Task SyncWrapper(Action sync)
        {
            sync();
            return Task.CompletedTask;
        }

        return SyncWrapper(action);
    }
}
```
and then you would treat everything as async internally.
I believe the only thing to deal with then is someone calling Fire() instead of Fire.Async but I'm not sure a consumer would define async delegates but then call the sync version of Fire. But I think this can just use a Task .Wait().

The complication I hadn't considered was how EvenCallback deals with old skool Tasks and returns TaskResult. I need to get my head around supporting the older versions of .net pre the async await days.

I would be happy to give this another go though.

Issue #453, opened 2021-06-03
CanFire does not support triggers with parameters


lloydjatkinson:
Currently `CanFire` has these two overloads:

```csharp
        public bool CanFire(TTrigger trigger, out ICollection<string> unmetGuards);
        public bool CanFire(TTrigger trigger);
```

However it does not support triggers with parameters, is there a technical limitation for this? Is there any possibility of adding support for this? It seems strange that I can check if the trigger can be fired for "normal" triggers and get a list of unmet guards, but I can't for triggers with parameters.

As I have a few normal triggers and a few parameterised triggers, the best way I can see for how to achieve checking if it can fire now is to wrap a call to `Fire` with a `try/catch`.

HenningNT:
I think the CanFire method should be made private, so that's it not available for users of the library.

My reasoning for this is that states are transient. Imagine calling CanFire, which involves checking some information in database. This information might change right after checking it, making the result of CanFire wrong.

I strongly suggest not depending on CanFire during run time. I suppose using it in unit tests is OK, since all information is static in a unit test.

If you use CanFire to make decisions on wether or not to Fire a trigger, I think you are not taking fully advantage of the state machine library.

If you use CanFire to drive a UI, then I suggest splitting your application up, so that you have one state machine dedicated just for the UI. You can hook up the transition events to drive the UI updates.

As to the question if there is a technical reason why CanFire can't accept pamaters, no there is none. It isn't implemented yet because nobody has asked for it yet :-)

lloydjatkinson:
I see, thanks 😊 So wrapping `Fire` with a `try/catch` to check if something was successful or not is your recommendation? I'm not checking whether to fire or not, I'm simply using `CanFire` currently to inform the consuming code if it was successful or not.

josellm:
I created a pull request about it.

https://github.com/dotnet-state-machine/stateless/pull/467

lloydjatkinson:
> You can hook up the transition events to drive the UI updates.

Did you mean for example, having two FSM's that communicate with each other? Not totally clear

tibitoth:
@HenningNT Firstly could you make `CurrentRepresentation` property `protected`? That could be enable more scenario for authors but not on the public API surface

SypleMatt:
+1 for a CanFire() method with arguments.

[...]

Issue #450, opened 2021-05-12
Check available transitions


anatoly-kryzhanovsky:
I need to have method which return true if transition with specified trigger and parameters available for current state.
because there are parameters i can't use CanFire, so i write following code:

```
public bool CanApply(TTriggers trigger, params object[] arguments)
{
	if(arguments == null || arguments.Length == 0)
		return StateMachine.CanFire(trigger);
	
	var permittedTriggers = StateMachine.GetPermittedTriggers(arguments);
	return permittedTriggers.Any(x => Equals(x, trigger));
}
```

and it works perfectly, but not always.

for example i have state machine with following transitions:
- new -> review, with string parameter (like reason)
- new -> onprogress with guid parameter (like document id)

now, then i call CanApply method with SendToWork as trigger and Guid as argument the ArgumentException was throwed. Stateless try to convert Guid to string.

i can't find any workaround to get available transition.

Environment:
- Platform: .net core 3.1
- Stateless: 5.11.0

anatoly-kryzhanovsky:
i create MVP for reproduce that case

```
enum Triggers
{
      SendToReview,
      SendToWork
  }

  enum States
  {
      New,
      Review,
      InProgress
  }

  class Entity
  {
      public States State { get; set; }
  }

  class Workflow
  {
      private readonly StateMachine<States,Triggers> _stateMachine;
      
      public bool CanApply(Triggers trigger, params object[] arguments)
      {
          if(arguments == null || arguments.Length == 0)
              return _stateMachine.CanFire(trigger);
	
          var permittedTriggers = _stateMachine.GetPermittedTriggers(arguments);
          return permittedTriggers.Any(x => Equals(x, trigger));
      }

      public Workflow(Entity entity)
      {
          _stateMachine = new StateMachine<States, Triggers>(() => entity.State, state => entity.State = state);
          var sendToReviewTrigger = _stateMachine.SetTriggerParameters<string>(Triggers.SendToReview);
          
          _stateMachine.Configure(States.New)
              .PermitIf(
                  sendToReviewTrigger,
                  States.Review,
                  _ => true);
      }
  }
  
  class Program
  {
      static void Main(string[] args)
      {
          var entity = new Entity
          {
              State = States.New
          };
          var workflow = new Workflow(entity);
          var canApply = workflow.CanApply(Triggers.SendToWork, Guid.NewGuid());
      }
  }
```

[...]

PR #454, opened 2021-06-07
[450] Surround unpack with try/catch block

Julio-Cam:
Addressing issues [450](https://github.com/dotnet-state-machine/stateless/issues/450) and [453](https://github.com/dotnet-state-machine/stateless/issues/453)

Triggers with parameters which contain different parameter types throw an exception when accessing the PermittedTriggers.


kuldeepGDI:
when will this get merged?

PR #457, opened 2021-08-15
Allow GetPermittedTriggers to Accept StronglyTyped Triggers with Guards in the Configurations

WahidBitar:
There is no need to throw an exception when trying to get the Permitted Triggers. If it's not valid just don't return it.
The previous behavior was throwing exceptions when you have different types of strongly typed triggers and you try to check the available triggers for one of them.
I've added a Test Case to cover the issue.

### DIFF
--- a/src/Stateless/StateRepresentation.cs
+++ b/src/Stateless/StateRepresentation.cs
@@ -318,12 +318,23 @@ namespace Stateless
 
             public IEnumerable<TTrigger> GetPermittedTriggers(params object[] args)
             {
-                var result = TriggerBehaviours
-                    .Where(t => t.Value.Any(a => !a.UnmetGuardConditions(args).Any()))
-                    .Select(t => t.Key);
+                var result = new HashSet<TTrigger>();
+                foreach (var triggerBehaviour in TriggerBehaviours)
+                {
+                    try
+                    {
+                        if (triggerBehaviour.Value.Any(a => !a.UnmetGuardConditions(args).Any()))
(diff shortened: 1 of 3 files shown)


anatoly-kryzhanovsky:
i dont think that exception flow is good for normal behaviour. 
exception is a slow mechanism, so we need to avoid them if we can

check my pull request - https://github.com/dotnet-state-machine/stateless/pull/451/files

WahidBitar:
@anatoly-kryzhanovsky  Thanks, I realize the performance issue about using exceptions. But I didn't want to change the logic in the internal implementation. I'm not quite sure about the internal implementation, are you sure that your changes will not affect the behavior of other places in the library?

anatoly-kryzhanovsky:
in old behavior in case of type mismatch exception will be throw and method will exit. in my code i added check for that case, no internal method was changed

WahidBitar:
@tlk May you check this, please

WahidBitar:
please!

crozone:
Looks good. This does remove the laziness of the `IEnumerable<TTrigger>` by building the `HashSet<TTrigger>` immediately but I don't think that matters much at all.

I guess it's just a question of whether we go with this option or https://github.com/dotnet-state-machine/stateless/pull/451/files

WahidBitar:
@mclift & @p-m-j  may you review this please.

p-m-j:
@WahidBitar - I haven't really looked at or thought about this project in years so I don't think I'm the right person to review, I don't think I ever really deserved membership of the org.

PR #467, opened 2021-11-10
- Method CanFire supports trigger with parameters

josellm:
I use CanFire for allow/deny actions(trigger) on our application's workflow. I need pass parameters to CanFire to determinate allowed actions in the current state.


ffMathy:
I have the possibility to merge this, but I am also new to this repo, so pardon my questions.

But isn't it possible to make a generic variant of this?

PR #470, opened 2022-03-07
Fix #449

### LINKED ISSUE #449: Change in name from Stateless.dll -> stateless.dll
Hi all.

I noticed when upgrading to version 5 that the filename changed from Stateless.dll to stateless.dll. This showed up on a Linux project which has a case-sensitive file system (it was running under Mono, not sure if that's a culprit?). Mono had problems resolving the dll until I changed the name back to a capital 'S'.

Anyhow, Mono isn't your problem :) But I wanted to ask if the name change was intentional and is planned to stay that way? If not, I'll keep my eyes out for that when I upgrade next.

Also, I'm not sure if the assembly resolution is a mono bug/problem or is something that could affect Stateless users on Linux using the `dotnet` runtime so I thought I'd include that info here in case it's helpful.

Thanks for all of your work on Stateless!

shaeussler:
Issue https://github.com/dotnet-state-machine/stateless/issues/449 fixed.


crozone:
Any word on getting this merged? #449 is causing build errors in Linux and Xamarin.

ffMathy:
Before I merge this, I have two questions.

Where is the filename changed in the diff? I only see a version bump. Also, how does this solve the problem without re-introducing it for people who already have a different filename?

crozone:
See my fresh comment here: https://github.com/dotnet-state-machine/stateless/issues/449#issuecomment-1225200256

There is definitely some toolchain weirdness when the github build calls `dotnet pack` with the lowercase `stateless.csproj`, which intermittently, and inexplicably causes the output .dll to also be lowercased as `stateless.dll`. You might want to fix that as well.

This PR just sets `<AssemblyName>Stateless</AssemblyName>`, which forces the build to always output the correct uppercased `Stateless.dll`, reglardless of what the project is called.

I'm assuming the people who are currently using the lowercased `stateless.dll` are simply in environments where it doesn't matter. It's still incorrect, but it's not causing issues due to different handling of case sensitivity, for whatever reason. Fixing this shouldn't negatively affect them, only correct the cases were it does (like Xamarin).

crozone:
More specifically, I think the original issue has to do with the `AssemblyTitle` being hardcoded to the uppercase `Stateless`, while the actual assembly name could be lowercase due to this weirdness/bug. This seems to confuse some toolchains like xamarin.

Setting both `AssemblyName` and `AssemblyTitle` to uppercase `Stateless` should solve the issue.

crozone:
@shaeussler You may want to rebase+squash this PR also...

crozone:
@ffMathy I have a single commit version of this PR that doesn't touch the version number here: https://github.com/dotnet-state-machine/stateless/pull/480

RaulSebastian (approved):

crozone:
Closing as it was fixed in https://github.com/dotnet-state-machine/stateless/pull/480

PR #474, opened 2022-05-27
Moved formating state to the base class

sulmar:
I moved the formatting initial state to the base class because the current solution blocks the possibility of adding a new graph format.


ffMathy:
I only see something added here, not something moved.

Can you elaborate more on that?

Issue #473, opened 2022-05-10
Support for .net 6?


ericsleung:
Will there be a new build to target .net 6 framework? Thanks

auvansang:
I'm waiting it too!

kerrpeter:
Can anyone confirm if the current version works in .Net 6 or not or what issues are expected?

alexeyshockov:
It works for me on .NET 6 without any issues

AlonAm:
Works great for me with NET MAUI on .NET 6, compiled together with my project. Waiting for the NuGet package 📦

PR #479, opened 2022-08-24
Add explicit .NET 6.0 framework support

crozone:
Add .NET 6.0 as a target framework, with the net6.0 TFM.

Corrected the .NET 5.0 TFM from net50 to net5.0

TFM documentation reference:
https://docs.microsoft.com/en-us/dotnet/standard/frameworks#supported-target-frameworks

### DIFF
--- a/src/Stateless/Stateless.csproj
+++ b/src/Stateless/Stateless.csproj
@@ -3,7 +3,7 @@
   <PropertyGroup>
     <AssemblyTitle>Stateless</AssemblyTitle>
     <Product>Stateless</Product>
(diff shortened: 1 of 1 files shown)


crozone:
Should close https://github.com/dotnet-state-machine/stateless/issues/473

crozone:
@ffMathy and a workflow check on this one too if it's not too much trouble

ffMathy:
There you go! 😊🙏

crozone:
Just discovered an issue with this PR: I forgot to also add the TFM changes to the testing project as well. I'll do that in a separate PR.

PR #480, opened 2022-08-24
Hardcode the AssemblyName attribute in the .csproj to "Stateless".

crozone:
This ensures that the output .dll is always named "Stateless.dll"
instead of "stateless.dll", which can occur due to quirks in the build
environment and the way the build system is called.

This, in turn, ensures that the output dll name aligns with the
AssemblyTitle attribute, ensuring that build systems consuming the
library are not confused.

This is a single commit version of the https://github.com/dotnet-state-machine/stateless/pull/470 PR with a more descriptive commit message. It also does not alter the version number. Should fix https://github.com/dotnet-state-machine/stateless/issues/449

### DIFF
--- a/src/Stateless/Stateless.csproj
+++ b/src/Stateless/Stateless.csproj
@@ -1,6 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
+    <AssemblyName>Stateless</AssemblyName>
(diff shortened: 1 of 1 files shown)


crozone:
@ffMathy Could I please get a workflow check on this one too?

ffMathy:
Yup! There you go! 😊

PR #484, opened 2022-09-21
Stateless.Tests -> .NET 6

crozone:
Update the Stateless.Tests project target framework from net50 to the net6.0 TFM, to re-target the tests project onto the latest .NET 6.0 LTS.


mclift:
Looks good and tests are green!

PR #488, opened 2022-10-05
Update Readme regarding state machine events

StevenTCramer:
Both `OnTransitioned`, and `OnTransitionedCompleted` take an Action that takes a single parameter `transition`.

Also `OnInternalTransitioned` doesn't exist so I removed the reference to it.


ffMathy:
Thanks! 😍

Issue #303, opened 2019-02-06
InternalTransitionAsyncIf vs InternalTransitionIf guard funcs


keirk:
Hi, 

I've spotted something that seems like a bug / oversight - 

`.InternalTransitionAsyncIf()` (when using a TriggerWithParameter) has a guard parameter signature of `Func<bool>`, where as `.InternalTransitionIf()` has a guard param signature of `Func<T, bool>`.  

It would be useful to receive the trigger parameter in the guard IMHO.

HenningNT:
That seems to be an oversight, yes… Would you like to fix it?

TheCollegedude:
I was trying to fix it, but it is not as easy as it seems, because the Async implementations drops all trigger parameters for the condition (only support `Func<bool> guard`)

#468

crozone:
I also just hit this. Will look into the changes required to fix it - ideally the triggers should also be passed into the guard.

Effectively, we need to align the async:

https://github.com/dotnet-state-machine/stateless/blob/05d44327bb33b511d6ad21de7f3fbe926e67b857/src/Stateless/StateConfiguration.Async.cs#L66-L73

with the sync:

https://github.com/dotnet-state-machine/stateless/blob/05d44327bb33b511d6ad21de7f3fbe926e67b857/src/Stateless/StateConfiguration.cs#L152-L159

PR #510, opened 2023-04-11
fix: InternalTransitionAsyncIf guard parameters

LipatovAlexander:
#303 

Should I keep the old "wrong" overloads for backwards compatibility?

### DIFF
--- a/src/Stateless/InternalTriggerBehaviour.cs
+++ b/src/Stateless/InternalTriggerBehaviour.cs
@@ -45,7 +45,7 @@ namespace Stateless
             {
                 readonly Func<Transition, object[], Task> InternalAction;
 
-                public Async(TTrigger trigger, Func<bool> guard,Func<Transition, object[], Task> internalAction, string guardDescription = null) : base(trigger, new TransitionGuard(guard, guardDescription))
+                public Async(TTrigger trigger, Func<object[], bool> guard,Func<Transition, object[], Task> internalAction, string guardDescription = null) : base(trigger, new TransitionGuard(guard, guardDescription))
                 {
                     InternalAction = internalAction;
                 }
--- a/src/Stateless/StateConfiguration.Async.cs
+++ b/src/Stateless/StateConfiguration.Async.cs
@@ -12,21 +12,6 @@ namespace Stateless
             /// <summary>
             /// Add an internal transition to the state machine. An internal action does not cause the Exit and Entry actions to be triggered, and does not change the state of the state machine
             /// </summary>
-            /// <param name="trigger"></param>
-            /// <param name="guard">Function that must return true in order for the trigger to be accepted.</param>
-            /// <param name="entryAction"></param>
-            /// <returns></returns>
-            public StateConfiguration InternalTransitionAsyncIf(TTrigger trigger, Func<bool> guard, Func<Transition, Task> entryAction)
-            {
-                if (entryAction == null) throw new ArgumentNullException(nameof(entryAction));
-
-                _representation.AddTriggerBehaviour(new InternalTriggerBehaviour.Async(trigger, guard, (t, args) => entryAction(t)));
-                return this;
-            }
-
-            /// <summary>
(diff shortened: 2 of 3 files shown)


crozone (approved):

mclift:
> Thanks for taking the time to fix this, it's much appreciated.
> 
> > Should I keep the old "wrong" overloads for backwards compatibility?
> 
> I'm thinking we should just take this as a breaking change, and do the appropriate version bump.
> 
> @mclift Thoughts?

Hi! Thanks for your contribution @LipatovAlexander!

@crozone, this seems like a fairly low-impact breaking change, and I notice that none of the existing unit tests broke as a result of this fix. Perhaps we could include something in the release note with a cheat sheet to updating code that's dependent on the old API.

LipatovAlexander:
> Perhaps we could include something in the release note with a cheat sheet to updating code that's dependent on the old API.

Maybe something like this:

If you are using `InternalTransitionAsyncIf` with a parameterized trigger, you should update your code to use `Func<TArg, bool>` instead of `Func<bool>` as a guard (in case of one parameter).

```csharp
var triggerX = stateMachine.SetTriggerParameters<string>(Trigger.X);
stateMachine.Configure(State.A)
    .InternalTransitionAsyncIf(triggerX, s => Guard(), Action) // in the old version it would be `() => Guard()`
```

mclift:
Having thought about it some more and having played around with the code in your branch, I think it would be better to retain the existing method overloads, mark them with `[Obsolete]`, and remove them at some point after the next major version.

```csharp
            [Obsolete("Use InternalTransitionAsyncIf(TTrigger, Func<bool>, Func<Task>) instead.")]
            public StateConfiguration InternalTransitionAsyncIf<TArg0>(TTrigger trigger, Func<bool> guard, Func<Transition, Task> internalAction) {...}

            ...

            [Obsolete("Use InternalTransitionAsyncIf<TArg0>(TriggerWithParameters<TArg0>, Func<TArg0, bool>, Func<TArg0, Transition, Task>) instead.")]
            public StateConfiguration InternalTransitionAsyncIf<TArg0>(TriggerWithParameters<TArg0> trigger, Func<bool> guard, Func<TArg0, Transition, Task> internalAction) {...}
```

The tests you added for the new method signatures can be easily adapted for the existing ones. This might include something like:

```csharp
        [Fact]
        [Obsolete]
        public async Task InternalTransitionAsyncIf_DeprecatedOverload_AllowGuardWithThreeParameters()
        {
            var trigger = sm.SetTriggerParameters<int, string, bool>(Trigger.X);
            const int intParam = 5;
            const string stringParam = "5";
            const bool boolParam = true;
            var guardInvoked = false;
            var callbackInvoked = false;

            sm.Configure(State.A)
                .InternalTransitionAsyncIf(trigger, () =>
                {
                    guardInvoked = true;
                    return true;
                }, (i, s, b, transition) =>
                {
                    callbackInvoked = true;
                    Assert.Equal(intParam, i);
                    Assert.Equal(stringParam, s);
                    Assert.Equal(boolParam, b);
                    return Task.CompletedTask;
                });

            await sm.FireAsync(trigger, intParam, stringParam, boolParam);

            Assert.True(guardInvoked);
            Assert.True(callbackInvoked);
        }

        [Fact]
        [Obsolete]
        public async Task InternalTransitionAsyncIf_DeprecatedOverload_GuardExecutedOnlyOnce()
        {
            var guardCalls = 0;
            var order = new Order
            {
                Status = OrderStatus.OrderPlaced,
                PaymentStatus = PaymentStatus.Pending,
            };
            var stateMachine = new StateMachine<OrderStatus, OrderStateTrigger>(order.Status);
            stateMachine.Configure(OrderStatus.OrderPlaced)
                 .InternalTransitionAsyncIf<int>(OrderStateTrigger.PaymentCompleted,
                       () => PreCondition(ref guardCalls),
                       _ => ChangePaymentState(order, PaymentStatus.Completed));

            await stateMachine.FireAsync(OrderStateTrigger.PaymentCompleted);

            Assert.Equal(1, guardCalls);
        }
```

Does this seem like a reasonable approach?

PR #514, opened 2023-04-12
StateMachineInfo.InitialState.Transitions throws if AddRelationships not called

leeoades:
For StateInfo, Substates, FixedTransitions and DynamicTransitions are only initialised by a call to AddRelationships.

However, for StateMachineInfo.InitialState, this never happens. 

Therefore StateMachineInfo.InitialState.Transitions throws a System.ArgumentNullException. 

The StateInfo object should not throw if "half initialised"

### DIFF
--- a/src/Stateless/Reflection/StateInfo.cs
+++ b/src/Stateless/Reflection/StateInfo.cs
@@ -143,7 +143,14 @@ namespace Stateless.Reflection
         /// <summary> 
         /// Transitions defined for this state.
         /// </summary>
-        public IEnumerable<TransitionInfo> Transitions { get { return FixedTransitions.Concat<TransitionInfo>(DynamicTransitions); } }
+        public IEnumerable<TransitionInfo> Transitions 
+        {
+            get { 
+                return FixedTransitions == null // A quick way to check if AddRelationships has been called.
+                            ? null 
+                            : FixedTransitions.Concat<TransitionInfo>(DynamicTransitions); 
+            } 
+        }
 
         /// <summary>
         /// Transitions defined for this state.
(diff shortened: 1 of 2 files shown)


src/Stateless/Reflection/StateInfo.cs
```diff
@@ -143,7 +143,14 @@ public class StateInfo
         /// <summary> 
         /// Transitions defined for this state.
         /// </summary>
-        public IEnumerable<TransitionInfo> Transitions { get { return FixedTransitions.Concat<TransitionInfo>(DynamicTransitions); } }
+        public IEnumerable<TransitionInfo> Transitions 
+        {
+            get { 
+                return FixedTransitions == null // A quick way to check if AddRelationships has been called.
+                            ? null
```
crozone, line 150: Why not use null coalescing like: `FixedTransitions?.Concat<TransitionInfo>(DynamicTransitions)`

leeoades: I've not contributed to this project before and was a little worried about what version of c# it was safe to use. My Rider kept suggesting all sorts of refactorings, pattern matching etc. So simply playing it safe.

crozone: Fair point, I believe it's C# 10 (the .NET 6.0 default) but there's no lang version set in the csproj. This is very readable so I think it's fine to leave as is.

mclift (approved):

mclift:
Thanks for your contribution, @leeoades!

leeoades:
I take it only a maintainer can merge the PR? @mclift

PR #518, opened 2023-04-24
Execute sync trigger directly

leeoades:
Instead of delegating the execution of the internal trigger behaviour to a thread pool thread and waiting for it, simply execute it.

Whilst Task.Run is recommended within libraries, it has a side effect of dropping the synchronization context.

This PR is part of a larger change that I am working on to fix an issue caused by the introduction of ConfigureAwait(false) everywhere (which also loses the synchronization context) and breaks queued mode usage in Microsoft Orleans. More on that to follow in a different PR.


leeoades:
Ah, actually, if the state machine has been triggered by a UI thread and the internal trigger is long-running, then the Task.Run would move the work off onto a Thread Pool thread. OK, I'll add this behaviour as part of my other change.

PR #520, opened 2023-04-25
Update example apps to net6.0

leeoades:
netcoreapp2.1 apps have a known vulnerability, which my IDE is keen to point out every time I load the stateless solution. I see little harm in updating them to net6.0?


mclift (approved):

mclift:
I've been thinking about some housekeeping tasks like this too. Thanks, @leeoades

PR #521, opened 2023-04-25
Fix typos

leeoades:
I highly recommend Respeller plugin if you haven't got it!


mclift (approved):
Thanks for taking the time to raise this PR!

PR #524, opened 2023-04-25
Remove obsolete TargetFrameworks

pentp:
Remove obsolete unsupported target frameworks: net40, netstandard1.0, net5.0. 
Replace net45 and net472 with net462 (lowest supported version of .NET Framework 4.x).


mclift:
Thanks @pentp 

I've also been giving this some thought lately, especially on the subject of whether or not to continue to build for netstandard1.0. As your PR shows, dropping support will allow for some simplifications in the codebase, so it's really a question as to whether anyone is likely to benefit from this project continuing to support it.

I hope you don't mind me creating a discussion for this as I feel it's worth asking for feedback from the wider community.

roozbehid:
Most libraries out there already dropped support for .netstandar 1.0 and .net40. So IMO this project has to drop it too in order to let it move on with new features. Most (all) new features wont be even provided in .net standard 2.0, so removing some of those legacy bags is a right choice IMO

mclift:
Nobody's raised any serious concerns about dropping support for the unsupported frameworks in the 6 weeks or so since this PR was raised, which seems like long enough. Thanks for your patience, @pentp!

Issue #527, opened 2023-04-29
Introduce FiringMode.Serial


gao-artur:
Currently, stateless supports two `FiringMode`'s: `Immediate` and `Queued`. Both of them have their use cases, but unfortunately, neither of them matches my needs.
`Immediate` will change the state even if another state change is in progress by another thread hence can be used only in a single-threaded environment.
`Queued` will enqueue the state change and return to the caller immediately, while the enqueued job will be executed by the thread which executes the current state change. Both immediate return and forwarding the work to another thread are unacceptable in my scenario. In addition, it has its own multi-threaded usage problems.
I suggest introducing a new `FiringMode.Serial`. In this mode, the call will block if there is an active state change and will fire only after it is finished. I'm just starting to learn the source code of the library, but it looks like this can be achieved very easily by introducing a `SemaphoreSlim` based lock in the `InternalFire(Async)` methods.
```c#
private readonly SemaphoreSlim _firingLock = new SemaphoreSlim(initialCount: 1, maxCount: 1);

void InternalFire(TTrigger trigger, params object[] args)
{
    switch (_firingMode)
    {
        ...
        case FiringMode.Serial:
            _firingLock.Wait();
            try
            {
                InternalFireOne(trigger, args);
            }
            finally
            {
                _firingLock.Release();
            }
            break;
        default:
           ...
    }
}

async Task InternalFireAsync(TTrigger trigger, params object[] args)
{
    switch (_firingMode)
    {
        ...
        case FiringMode.Serial:
            await _firingLock.WaitAsync();
            try
            {
                await InternalFireOneAsync(trigger, args);
            }
            finally
            {
                _firingLock.Release();
            }
            break;
        default:
            ...
    }
}
```

Also, since the lock can be held for a long time, it's good to introduce `Fire(Async)` overloads that accept `CancellationToken`.

leeoades:
I didn't think stateless supported being used in a multithreaded way. 

Having multiple threads attempting state changes seems extremely problematic. One thread could be checking a guard whilst another thread changes the state immediately. The first thread then approves the state change and moves to the target state which may not be a permittable transition.

I wouldn't be encouraging users to interact with a state machine in a multithreaded fashion. If I were you, I would be looking to queue the transitions and have them processed sequentially.

gao-artur:
[...]

PR #535, opened 2023-07-03
Fixed a null reference exception issue in InternalFireQueuedAsync method and added a fix to support multi-tasks firing triggers

YaeliGimelshtein:
Using this NuGet in my own app, I came across a null reference exception in InternalFireQueuedAsync  method.
After a bit of investigation I came to a conclusion that there are 2 things needed to be fixed:
1) there is a use of object Queue when it's better to use ConcurrentQueue in order to be thread safe and avoid exceptions.
2) there was no lock in the InternalFireQueuedAsync  method so , what could happen (and happened in my own app) is that two threads will be able to go at the same time into the Dequeue sections and one of them will get null (if for example the queue has only one element).

I created 3 fixes:
1) changed the Queue object to ConcurrentQueue to support multiple threads
2) fixed the logic in the InternalFireQueuedAsync  to support double-checked locking in order to make sure that threads will not be able to dequeue at the same time.
3) created a new test to test this functionality 

for any more questions please feel free to contact me: yaeli.gimelshtein@kornit.com

### DIFF
--- a/src/Stateless/StateMachine.Async.cs
+++ b/src/Stateless/StateMachine.Async.cs
@@ -135,28 +135,37 @@ namespace Stateless
         /// <param name="args">     A variable-length parameters list containing arguments. </param>
         async Task InternalFireQueuedAsync(TTrigger trigger, params object[] args)
         {
-            if (_firing)
+            if (!_firing)
             {
-                _eventQueue.Enqueue(new QueuedTrigger { Trigger = trigger, Args = args });
-                return;
-            }
+                await _lock.WaitAsync();
 
-            try
-            {
-                _firing = true;
-
-                await InternalFireOneAsync(trigger, args).ConfigureAwait(RetainSynchronizationContext);
-
-                while (_eventQueue.Count != 0)
+                if(!_firing)
                 {
-                    var queuedEvent = _eventQueue.Dequeue();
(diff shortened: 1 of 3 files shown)


vlm---:
Stateless is not designed to be thread safe as is noted in [https://github.com/dotnet-state-machine/stateless/blob/dev/README.md#async-triggers](Readme). You can find some closed issues proposing similar fixes, but its easier for the user to ensure that the state machine is not used concurrently.

crozone:
Thanks for the PR, but we need to have a design discussion surrounding the design of a threadsafe mode for Stateless since the current queue modes aren't intended to be threadsafe at all.

Please see https://github.com/dotnet-state-machine/stateless/issues/527 for discussion surrounding adding `FiringMode.Serial` for concurrent event firing. We are considering a `SemaphoreSlim` as the queuing mechanism since it will also support the async usecases.

YaeliGimelshtein:
Thanks for your fast response. That's really weird since we have been using stateless for almost 2 years and for most of the time we never ran into a problem regarding parallelism or threads issues.  Hoping you consider this since most apps today use concurrency. thanks :)

YaeliGimelshtein:
Also, as in the mentioned PR, I also suggest the use of `SemaphoreSlim` but in a bit different way, since there is no need to add a new mode (queued will work perfectly with my change in order to support concurrency)

PR #536, opened 2023-07-14
Added FireAsync(TriggerWithParameters, params object[]) overload

DeepakParamkusam:
This is async counterpart to existing Fire(TriggerWithParameters, params object[]) method. Required for firing triggers with more than 3 parameters asynchronously.


DeepakParamkusam:
@mclift Could you check this PR and add it to 5.14 release if all is ok?

mclift:
Thanks for your contribution, @DeepakParamkusam

Issue #543, opened 2023-09-22
Using PermitIf on a state with substates seems to lead to reentry


hooligan495:
Hi, 

I've a setup where I have a state with Substates. 

When I configure the state with a permitif trigger guard, I fire the trigger and the substate transitions..  but if the trigger fires again.. the entry action of the substate I'm already in fires.  This seems counter to the documentation I've been reading.  

Is this supposed to happen? (I don't want the onentry to be called multiple times).   or is this a missunderstanding on my part. 

So say I configure a State with a parameterized trigger:

`
.Configure(SuperStateA).PermitIf(_paramTrigger, SubstateA, p == Event1)
.Configure(SubstateA).SubstateOf(SuperStateA).OnEntry(()=>{log.Info("HI)});
`

If my Event 1 fires multiple times I see Hi printed multiple times. 

If this is how you would expect Stateless to behave can someone indicate so?  I will update my guard conditions to account for the current state... 

Thanks

HenningNT:
I think that this should not have happened.

If the state machine is in SubstateA, then the trigger should be handled by the super state handler. So far, so good.
But the state doesn't change, so the state machine should not execute the onEntry action.

So I would say this is a bug.

hooligan495:
Thanks

hooligan495:
@HenningNT 
in an attempt to be helpful I did create this unit test in the Stateless solution.  and implemented the fix this way.. .(I'm not real happy with this fix as it seems to go against the conventions this project has been following (meaning I feel like finding the trigger should determine that it gets ignored vs what I did)).

If you have other suggestions on how this should be addressed I'd be happy to modify my changes and open a PR. 
If you like this and would like me to open a PR I'm fine doing that too. 


Unit test: 
`
[Fact]
public void WhenInSubstate_TriggerSuperStateTwiceToSameSubstate_DoesNotReenterSubstate()
{
    var sm = new StateMachine<State, Trigger>(State.A);
    var eCount = 0;

    sm.Configure(State.B)
        .OnEntry(() => { eCount++;})
        .SubstateOf(State.C);

    sm.Configure(State.A)
        .SubstateOf(State.C);


    sm.Configure(State.C)
        .Permit(Trigger.X, State.B);

    sm.Fire(Trigger.X);
    sm.Fire(Trigger.X);

    Assert.Equal(1, eCount);
}
`

code change in statemachine.cs
`
[...]

PR #544, opened 2023-10-05
In the case of a superstate having a trigger that moves the state to …

hooligan495:
…a substate, if the machine is already in that substate then the entry action shouldn't trigger again. this pr is for Issue #543


hooligan495:
If there is a better approach let's discuss.  After going through the stateless source code tho, this approach seemed the most stable to me.  but this is my first foray in to stateless.

HenningNT (approved):

HenningNT:
Look very good to me!
Thank you for your contrubution!

crozone:
Looks good to me, if I understand the case correctly the intentional re-entry case will be handled by `InternalTriggerBehaviour`, so just stops the superstate edgecase where the superstate re-enters itself in the `TransitioningTriggerBehaviour` case.

crozone (approved):

PR #551, opened 2023-11-30
Update to .NET 8

ArturKarbone:


mclift:
Thanks, @ArturKarbone.

Looks good to me, and the [guidance on target frameworks](https://learn.microsoft.com/en-us/dotnet/standard/library-guidance/cross-platform-targeting#net-and-net-standard-targets) seems to confirm there shouldn't be any compatibility issues for .NET 6.0 projects.

Issue #557, opened 2023-12-20
Remove dependency from dirgraph format


sulmar:
Hi,
I'm working on exporting the state machine to the **mermaid format**.

I have currently like this:

``` csharp
public class MermaidGraphStyle : GraphStyleBase
{
    public override string GetPrefix()
    {
        return "stateDiagram-v2";
    }

  // ...

``` 

and unit test:

``` csharp
// Arrange
StateMachine<string, string> _machine = new StateMachine<string, string>("s1");

_machine.Configure("s1")
    .Permit("A transition", "s2");

var machineInfo = _machine.GetInfo();

// Act
var result = Stateless.Graph.MermaidGraph.Format(machineInfo);

// Assert
string expected = "stateDiagram-v2
	s1 --> s2 : A transition";

Assert.AreEqual(expected, result);
 
```

In result I get:

``` 
stateDiagram-v2
	s1 --> s2 : A transition
 init [label="", shape=point];
 init -> "s1"[style = "solid"]
}
```

The problem is because in _ToGraph()_ method there is a strong dependence on the digraph format:

 ``` csharp
public string ToGraph(GraphStyleBase style)
{
 ...
 // Add initial transition if present
 var initialStateName = initialState.UnderlyingState.ToString();
 dirgraphText += System.Environment.NewLine + $" init [label=\"\", shape=point];";
 dirgraphText += System.Environment.NewLine + $" init -> \"{initialStateName}\"[style = \"solid\"]";

 dirgraphText += System.Environment.NewLine + "}";

...
}

```

Could you please remove in Stata Graph.To Graph() method dependency from dirgraph format?

Maybe good idea is refactor this section into abstract method in GraphStyleBase class?

sulmar:
I just created pull request #558

mclift:
Resolved in release 5.15.0

PR #558, opened 2023-12-20
Refactor Add initial transition to GetInitialTransition method

sulmar:
It's my proposal for change based on #557

This will open up new possibilities for defining your own formatters.

### DIFF
--- a/src/Stateless/Graph/UmlDotGraphStyle.cs
+++ b/src/Stateless/Graph/UmlDotGraphStyle.cs
@@ -1,4 +1,5 @@
-﻿using System;
+﻿using Stateless.Reflection;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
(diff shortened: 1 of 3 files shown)


sulmar:
I just created a branch on which I started testing the export to Mermaid diagrams:
https://github.com/sulmar/stateless/blob/add-mermaid-graph-style

Still needs refinement but looks like it's starting to work after refactoring.

mclift (commented):
Thanks, @sulmar

The new `abstract` method in `GraphStyleBase` will be a breaking change for existing derived classes, albeit a trivial one to fix. I've been considering whether adding it as a `virtual` method with a default implementation might be worth considering, but I don't imagine this change will affect many consumers of Stateless and creating a default implementation for backwards compatibility would just be creating technical debt and kicking the inevitable change further into the future. I'd be interested to know if others have any views on this.

src/Stateless/Graph/UmlDotGraphStyle.cs
```diff
@@ -122,5 +123,21 @@ internal string FormatOneLine(string fromNodeName, string toNodeName, string lab
         {
             return $"\"{fromNodeName}\" -> \"{toNodeName}\" [style=\"solid\", label=\"{label}\"];";
         }
+
+        /// <summary>
+        ///
```
mclift, line 128: Missing summary (I think you could replace this with `/// <inheritdoc/>`)

mclift:
3 days and no additional comments, so let's get this merged. Thanks, @sulmar

PR #559, opened 2023-12-20
Add mermaid graph style

sulmar:


mclift:
@sulmar Thanks for starting the work on this. I'll follow up soon with a PR to make it more complete, but please let me know if you're already working on it.

sulmar:
@mclift Yes, before I asked you for refactoring, I created an independent project: https://github.com/sulmar/stateless-mermaid
I would now like to move this solution to your library. Please let me know when is the best time to do this.

mclift:
The code in sulmar/stateless-mermaid seems very close to the changes in your pull request. Would you mind if I built on your work to improve feature parity with the Graphviz style and add test coverage? I'd do this in the new feature branch and create a PR from there.

sulmar:
Yes, it is very similar because the pull request was created based on this project. Sure, grab my code. I will be very happy because the stateless library is one of my favorites. I have given many presentations on it. For example on the [MS Tech Summit 2022](https://github.com/sulmar/mstechsummit2022-state-machine).

I also have another idea, but I'll work on it after your commit.

mclift:
That's really cool! Okay, to be continued...

PR #566, opened 2024-02-04
Small fixes: Potential NullReferenceException in the InvocationInfo class and FireAsync_TriggerWithMoreThanThreeParameters test failure

PolarGoose:
* Fix a potential NullReferenceException in the InvocationInfo class
* Make FireAsync_TriggerWithMoreThanThreeParameters unit test culturally invariant. Otherwise, it fails on some system locales.


mclift:
Thanks for your contribution, @PolarGoose!

Issue #565, opened 2024-01-26
Problem with Reentry in PermitDynamic in 5.14


Fieslix:
It seems like since 5.15 PermitDynamic is silently not re-entering the state.

        private enum State {
            Ready,
        }

        private enum Trigger {
            Info
        }

        [Test]
        public void DynamicEntry() {
            var stm = new StateMachine<State, Trigger>(State.Ready);
            var entryFromInfo = false;
            var entry = false;
            var config = stm.Configure(State.Ready);
            config.PermitDynamic(Trigger.Info, () => State.Ready);
            config.OnEntry(() => entry = true);
            config.OnEntryFrom(Trigger.Info, x => entryFromInfo = true);
            stm.Fire(Trigger.Info);
            Assert.That(entry, Is.True, "Entered into Ready");
            Assert.That(entryFromInfo, Is.True, "Entered from trigger info");
        }

DeepakParamkusam:
This is due to change made in #544. State is no longer re-entered unless the trigger is explicitly configured via `PermitReentry`. `PermitDynamic` is also affected due to this.

However, if `config.PermitReentry(Trigger.Info);` is added to above test configuration, the state machine throws `InvalidOperationException`-
```
System.InvalidOperationException: 'Multiple permitted exit transitions are configured from state 'Ready' for trigger 'Info'. Guard clauses must be mutually exclusive.'
```

It appears that it is no longer possible to re-enter a state dynamically, which I believe is a regression.

@mclift , @crozone , @HenningNT Could you take a look?

mclift:
Thanks for raising this, @Fieslix & @DeepakParamkusam.

As I see it, allowing state re-entry through a dynamic transition should never have been the default behavior as it's inconsistent with the non-dynamic transitions and makes the API confusing. I also agree that there should be a way to allow a dynamic transition to re-enter the source state.

As `PermitDynamic` and `PermitDynamicIf` already have 20+ method signatures between them, perhaps an additional parameter with a default value would be appropriate here.

To represent this internally, perhaps there's an argument for replacing the `ReentryTriggerBehaviour` class with a property on `TriggerBehaviour` that sets whether reentry is enabled in instances of its implementations.

So we might end up with something like:

```csharp
config.PermitDynamic(Trigger.X, () => State.A, permitReentry: true);
```

How does this sound as an approach?

Fieslix:
This is a difficult design decision.
[...]

PR #571, opened 2024-04-19
Permit state reentry from dynamic transitions.

mclift:
To address issue #565, this PR restores the pre-v5.15 behavior so that state re-entry is possible when dynamic transitions are used.

Example:

```csharp
sm.Configure(State.A).PermitDynamic(Trigger.X, () => State.A);
Console.WriteLine($"Initial state: {sm.State}");

// Firing Trigger.X does not cause InvalidOperation to be thrown.
sm.Fire(Trigger.X);
Console.WriteLine($"Final state: {sm.State}");

// Output:
// Initial state: A
// Final state: A
```

### DIFF
--- a/test/Stateless.Tests/AsyncActionsFixture.cs
+++ b/test/Stateless.Tests/AsyncActionsFixture.cs
@@ -4,7 +4,6 @@ using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
-using System.Threading;
 using System.Threading.Tasks;
 using Xunit;
 using System.Globalization;
@@ -562,7 +561,6 @@ namespace Stateless.Tests
                 });
 
             await sm.FireAsync(Trigger.X, 42, "Stateless", true, 123.45, Trigger.Y);
-            Console.WriteLine(Thread.CurrentThread.CurrentCulture);
 
             Assert.Equal(expectedParam, actualParam);
         }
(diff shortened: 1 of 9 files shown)


test/Stateless.Tests/AsyncActionsFixture.cs
```diff
@@ -559,15 +559,15 @@ public async Task FireAsyncTriggerWithParametersArray()
...
+            await sm.FireAsync(Trigger.X, 42, "Stateless", true, 123.45, Trigger.Y);
 
             Assert.Equal(expectedParam, actualParam);
         }
 
         [Fact]
         public async Task FireAsync_TriggerWithMoreThanThreeParameters()
         {
-            const string expectedParam = "42-Stateless-True-420.69-Y";
+            const string expectedParam = "42-Stateless-True-123.45-Y";
```
Fieslix, line 570: The behavior of ToString is based on the current user's locale setting.
This test will fail on machines with a different locale.
One could either use ToString in the input as well or pin the locale.
Thread.CurrentThread.CurrentCulture = System.Globalization.CultureInfo.InvariantCulture;
string expectedParam = $"42-Stateless-True-{123.45}-Y";

mclift: Good spot, @Fieslix, thank you. Should be resolved now.

Fieslix (approved):

Issue #550, opened 2023-11-29
Handling of value based trigger parameters 


gartenriese2:
When using value based trigger parameters, CanFire returns true by default. For example:

`var valueTrigger = stateMachine.SetTriggerParameters<int>(Trigger.X);`
`stateMachine.Configure(State.Y).PermitIf(valueTrigger, State.Z, i => i < 5);`
`Assert.IsFalse(stateMachine.CanFire(Trigger.X));`

The assertion fails because when no parameter is used for a trigger, the default value for the parameter type is used in PermitIf.

PR #574, opened 2024-05-12
Add overloads to CanFire to support trigger parameters

mclift:
To address #550

This PR adds support for trigger parameters to `CanFire`. Example:

```csharp
const int argument0 = 101;
var sm = new StateMachine<string, string>("StateA");
var valueTrigger = sm.SetTriggerParameters<int>("MyTrigger");
sm.Configure("StateA").PermitIf(valueTrigger, "StateB", i => i > 100);

Console.WriteLine($"CanFire with {valueTrigger.Trigger} " +
    $"on state {sm.State} " +
    $"with parameter value {argument0} " +
    $"returns: {sm.CanFire(valueTrigger, argument0)}.");

// Output:
// CanFire with MyTrigger on state StateA with parameter value 101 returns: True.
```

This PR does not change the existing behavior of `bool CanFire(TTrigger)` nor of `bool CanFire(TTrigger, out ICollection<string>)`, but xmldoc remarks have been added to clarify that they may not work as expected when used with parameterized triggers.

### DIFF
--- a/src/Stateless/StateMachine.cs
+++ b/src/Stateless/StateMachine.cs
@@ -570,6 +570,12 @@ namespace Stateless
         /// Returns true if <paramref name="trigger"/> can be fired
         /// in the current state.
         /// </summary>
+        /// <remarks>
+        /// When the trigger is configured with parameters, the <c>default</c> value of each of the trigger parameter's types will be used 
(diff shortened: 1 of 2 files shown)


crozone:
LGTM!

PR #575, opened 2024-05-21
Remove getDestination, and use Destination property instead.

HenningNT:
Remove getDestination, and use Destination property instead.
Only dynamic transition needs a GetDestination, so it still has it. Updated tests according to changes.


mclift (approved):
Looks like a good optimisation 👍

PR #581, opened 2024-05-31
Expose unmet guard conditions

andrewmckaskill:
This change adds a method to expose the unmet guard conditions.

I'm using this to show validation messages to users as to WHY they are unable to take a certain action within the workflow.

While it is possible to get the entire set of guard condition descriptions through the `StateMachine.GetInfo()` method, the info object exposes all guard conditions for a Trigger, without reporting back which ones are unmet. 

Please let me know if this needs additional changes, and what kind of tests would be appropriate.

I'm also happy to change the return value `IEnumerable<Tuple<TTrigger, string[]>>` with `Dictionary<TTrigger, ICollection<string[]>` if that more closely aligns with other types in use - please let me know.

### DIFF
--- a/src/Stateless/StateMachine.cs
+++ b/src/Stateless/StateMachine.cs
@@ -147,6 +147,17 @@ namespace Stateless
                 .Select(trigger => new TriggerDetails<TState, TTrigger>(trigger, _triggerConfiguration));
         }
 
+        /// <summary>
+        /// Gets any triggers for the current state that are not permissible, and the descriptions
+        /// of the unmet trigger conditions associated with them.
+        /// </summary>
+        /// <param name="args"></param>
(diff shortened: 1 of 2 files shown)


mclift:
Thanks for taking the time to contribute!

From what you've described, it looks like the overloads of `CanFire` with `out ICollection<string> unmetGuards` are a close fit, although they differ in that you need to provide the trigger you want to test. Would `CanFire` meet your needs here?

Mike

andrewmckaskill:
Thanks for the feedback Mike. I've had a review of the code and yes, `CanFire` does meet my needs. I thought I had scoured the codebase thoroughly, but I guess I missed it. I'll go ahead and cancel this PR. 

Thanks!

PR #583, opened 2024-06-14
Use license expression

lg2de:
https://learn.microsoft.com/en-us/nuget/create-packages/package-authoring-best-practices#licensing

It is recommended to use the `PackageLicenseExpression`.


mclift:
Thanks @lg2de, well spotted and thank you for the PR!

lg2de:
A colleague has just pointed out to me that we both, @mclift , have overlooked the fact that PackageLicenseExpression is now a duplicate. Oops!
I'll create another PR to fix that.

PR #584, opened 2024-06-14
Remove duplicate PackageLicenseExpression

lg2de:
I accidentally created a duplicate in #583 .


mclift:
Ah, thanks for the follow-up!

PR #595, opened 2024-07-16
Allow permitDynamic destination state to be calculated with an async function (Task)

Jon2G:


mclift:
Thanks for taking the time to contribute, @Jon2G 

As there's no issue associated with this PR, I'm curious to know if this addresses a particular problem you've encountered or if it's intended purely to bring parity between the synchronous and the `async` configuration methods.

Some feedback on the unit tests (thank you for writing them!):
- Some of the new methods don't appear to have test coverage; it would be good to make sure they all have tests.
- Many of the tests include a short delay that doesn't seem necessary; could they return immediately? Something like `.PermitDynamicAsync(Trigger.X, async () => await Task.FromResult(State.B))` perhaps?

mclift (approved):

jornhd:
Thanks! I just needed that too. Currently using `.GetAwaiter().GetResult()` as a workaround.

PR #597, opened 2024-08-04
Escape labels in `UmlDotGraphStyle`

nightroman:
`UmlDotGraph` does not escape special characters like `"` and `\` in state,
trigger, description labels. As a result, the generated DOT is invalid when
such labels exist.

This is not a big deal perhaps when labels are carefully selected in order to
work around this issue. But state machines may be dynamically created from
data/strings not originally designed with `Stateless` and this issue in mind.

This PR includes:

- new method `UmlDotGraphStyle.EscapeLabel`
- `UmlDotGraphStyle` uses this method when required
- added the new test `SimpleTransitionWithEscaping`, similar to `SimpleTransition` but using problematic labels
- removed the test `SimpleTransitionUML` because it seems to be completely identical to `SimpleTransition`
- reworked the existing test `SpacedUmlWithSubstate` so that it uses problematic states, triggers, descriptions

P.S. I hit this issue with real use cases on making state machines in PowerShell for using with [FarNet.Stateless](https://github.com/nightroman/FarNet.Stateless).

### DIFF
--- a/src/Stateless/Graph/UmlDotGraphStyle.cs
+++ b/src/Stateless/Graph/UmlDotGraphStyle.cs
@@ -29,21 +29,20 @@ namespace Stateless.Graph
         public override string FormatOneCluster(SuperState stateInfo)
         {
             string stateRepresentationString = "";
-            var sourceName = stateInfo.StateName;
 
-            StringBuilder label = new StringBuilder($"{sourceName}");
+            StringBuilder label = new StringBuilder($"{EscapeLabel(stateInfo.StateName)}");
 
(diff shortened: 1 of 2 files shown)


crozone:
Hi @nightroman, just getting around to reviewing this, thanks for the bump. It might take a day or two.

nightroman:
bump

mclift:
Thanks for this, @nightroman.

nightroman:
@mclift Thank you, looking forward to the NuGet package update!

PR #606, opened 2024-12-02
Created a custom awaiter that guarantees that we complete on a different threadpool thread. 

leeoades:
Previously, an awaited Task had a non-zero chance of completing before it was awaited, in which case execution would continue on the same thread. I created a unit test that repeatedly attempted to lose the context in order to detect race conditions which proved that other techniques (using Task) simply reduced the probability instead of guaranteeing success.


leeoades:
Rebased onto https://github.com/dotnet-state-machine/stateless/pull/608 in order to remove the dodgy dependencies.

crozone (approved):

PR #617, opened 2025-04-01
Processing #Task macro related code

SaNeOr:


mclift:
I've just removed the `#if TASK` directives in another PR as `async` is supported in all current target frameworks. Have you encountered a scenario where these are still needed?

PR #618, opened 2025-07-01
PermitIfAsync and PermitReentryIfAsync support

ryangomesvr:
Added support for PermitIfAsync and PermitReentryIfAsync, following on from discussions as part of https://github.com/dotnet-state-machine/stateless/issues/189

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,12 @@ phoneCall.Configure(State.OffHook)
     .PermitIf(Trigger.CallDialled, State.Beeping, () => !IsValidNumber);
 ```
 
+```csharp
+phoneCall.Configure(State.OffHook)
+    .PermitIfAsync(Trigger.CallDialled, State.Ringing, async () => await IsValidNumber())
+    .PermitIfAsync(Trigger.CallDialled, State.Beeping, async () => !await IsValidNumber());
+```
+
 Guard clauses within a state must be mutually exclusive (multiple guard clauses cannot be valid at the same time.) Substates can override transitions by respecifying them, however substates cannot disallow transitions that are allowed by the superstate.
 
 The guard clauses will be evaluated whenever a trigger is fired. Guards should therefore be made side effect free.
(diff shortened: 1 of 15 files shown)


ryangomesvr:
> @ryangomesvr it looks like a lot of thought has gone into this. Thanks for creating this PR!
> 
> As it is, this will be a breaking change for existing callers of `PermittedTriggers`, `GetPermittedTriggers` and `GetDetailedPermittedTriggers`, albeit just a change to an `async` call. Is there something driving this choice that I've missed, versus creating `async` versions of them?

@mclift thank you for your question! My thinking was based purely on the fact that someone using a version of the package containing this change would require a very minor change to support it. 

Just keeping the existing GetPermittedTriggers in StateRepresentation would potentially be problematic as if a user of the package were to call PermitIfAsync with an async guard, it would not be picked up by the existing GetPermittedTriggers call but would still look as if everything runs as expected, so this change was to avoid it.

I am certainly not against the idea of keeping a synchronous version of the above methods and having an async alternative separately, but wanted to avoid wrapping the call to GetPermittedTriggersAsync in a Task.Run and calling .GetAwaiter().GetResult().

Would it be preferable to do as such to maintain support for existing users and still getting support for the asynchronous permitted checks, or am certainly open to any other ideas as well?

mclift:
@ryangomesvr Instinctively I feel it's better to maintain compatibility as long as it doesn't cause more problems, and to consider using `ObsoleteAttribute` to give users some warning if there's an intent to remove something they might be using.

ryangomesvr:
@mclift Thank you for your feedback. 

I have updated the PR to bring back the original PermittedTriggers/GetPermittedTriggers marked as obsolete and created async overrides for these, along with unit tests to ensure the legacy and new triggers have the same desired effect.

mclift:
Thanks for doing that, @ryangomesvr! This PR addresses a long-standing gap in the API and I'm sure it'll benefit a lot of people.

I'll try to get a release ready in the next few days, unless another maintainer gets there first :)

Issue #619, opened 2025-07-14
OnTransitioned() memory-leak, unable to unregistered from OnTransitionedEvent


jBijsterboschNL:
When calling the OnTransitioned() on the StateMachine, a registration (event handler) is being added on the _onTransitionedEvent. But there seems to be no option to unregistered from this event. This results in a memory leak.

RedOrc:
@jBijsterboschNL Funny that you mentioned it because I was perusing the code and saw the same anomaly.. I will take a look at it later on and will create a patch to address it. Should have a PR up by tomorrow the latest 😉 .

RedOrc:
@jBijsterboschNL I've got the code done but I need to update a few of the examples and documentation. Will post details of the fix once applied.

mclift:
@jBijsterboschNL It's a relatively small change to provide methods to unregister event handlers (thanks @RedOrc for #625), but can you describe where the memory leak is coming from? Is your application dynamically registering new event handlers during execution—and therefore growing the collection of event handlers indefinitely—or is there something else going on? Is it something you can show with a unit test or sample program? I'm just keen to make sure we're addressing the right issue.

RedOrc:
@mclift I think that @jBijsterboschNL may be quite busy so I'll jump in and add a few scenarios that come to mind.

* I am imagining a scenario in which I have a GUI application and I am registering to the state machine in order to receive transition events so I can automatically switch screens when the events are raised. However, I only want to do this for a short period of time while a particular GUI is active. In the case that the user continuously opens and closes the GUI window in a long running process, the inability to unregister would slowly leak memory as a live reference is now held and unable to be reclaimed by the garbage collector.

* Prior to my PR, double subscription was possible which meant that the developer could unknowingly add multiple callbacks of the same object reference. With no way to unregister the callbacks, the garbage collector cannot reclaim that memory or that of the subscribed class object holding the reference.

*  Subscribing to events in a Web API service and logging the various transitions for some audit related purpose. Same problem as the two previous scenarios, but this time at scale. Classic example is shown in this [Medium Article](https://m3lles.medium.com/stateless-a-lightweight-workflow-library-alternative-for-net-b739389bb049) on someone using this library in their API service 🧑‍🚒 .

[...]

PR #625, opened 2025-08-21
Issue619 - Provide method to unregister events

RedOrc:
# Problem

In [Issue 619](https://github.com/dotnet-state-machine/stateless/issues/619) it was reported by a user that there was no way to unregister events from the state machine. This leads to a memory leak in long running applications if you no longer want to be notified of ```transition``` or ```transition completed``` events.

# Solution

Provide a way to unregister from both synchronous and asynchronous events. To successfully unregister callbacks, callers must maintain a reference to the original callback to ensure that it is cleared up. Otherwise, they can simply call ```stateMachine.UnregisterAllCallbacks()``` in order to remove all callbacks (no reference needed).

Documentation has been updated to reflect these new methods and the ```OnOffExample``` program has been tweaked to demonstrate registering and unregistering from callback events.

# Changes

* Updated documentation to reflect new state machine API methods.
* Updated ```OnOffExample``` program to demonstrate the usage of the added APIs in the state machine.
* Added safety guards to prevent double subscription to the same callback method in both synchronous and asynchronous workflows.
* Added ability to Unregister from both synchronous and asynchronous workflows. Additional ```UnregisterAll``` method has been added to remove all callbacks at once with no reference pointer needed. 
* Added new ```OnTransitionedEventTest.cs``` class to specifically test the ```OnTransitionedEvent.cs``` class. These tests provide coverage for the entire functionality of the transition event class.


RedOrc:
@mclift I've got another one for you 🤠. It's a fix for the aforementioned issue in my description.

test/Stateless.Tests/OnTransitionedEventTest.cs
mclift, line 1: Should probably rename to `OnTransitionedEventTests.cs` to be consistent with other test classes using this naming convention.

RedOrc: @mclift Fixed 👍

mclift:
Thanks for preparing this PR!

Issue #624, opened 2025-08-20
Breaking change in behaviour with 5.18.0?


MattVane:
I'm not sure if this is intended behaviour or not, but I have noticed a change in behaviour between v5.17.0 and v5.18.0 after upgrading recently.
In the situation where a trigger transition is defined in a superstate and then overridden in a substate, I expected the substate transition to take precedence, as per the docs. This is definitely how v5.17.0 worked.
Since the refactor for new _PermitIfAsync_ method, the substate transition does not happen when _FireAsync()_ is called e.g.

```csharp
public class Stateless_v5_18Test
{
    private readonly StateMachine<MentalState, MentalTrigger> _sm = new(MentalState.Awake);

    public StatelessTest()
    {
        _sm
            .Configure(MentalState.Awake)
            .Permit(MentalTrigger.YouShouldBeWorking, MentalState.Working);
        // Overrides the superstate transition
        _sm
            .Configure(MentalState.Working)
            .SubstateOf(MentalState.Awake)
            .Permit(MentalTrigger.YouShouldBeWorking, MentalState.WorkingHarder);
    }

    private enum MentalState
    {
        Awake,
        Working,
        WorkingHarder,
    }

    private enum MentalTrigger
    {
        YouShouldBeWorking,
    }

    [Fact]
    public async Task GivenTriggerHandledOnSuperStateAndSubState_WhenTriggerFiredAsync_ThenShouldUseSubstateTransition()
    {
        await _sm.FireAsync(MentalTrigger.YouShouldBeWorking);
        Assert.Equal(MentalState.Working, _sm.State);

        await _sm.FireAsync(MentalTrigger.YouShouldBeWorking);
        Assert.Equal(MentalState.WorkingHarder, _sm.State); // FAILS!
    }

    [Fact]
    public void GivenTriggerHandledOnSuperStateAndSubState_WhenTriggerFiredSync_ThenShouldUseSubstateTransition()
    {
        _sm.Fire(MentalTrigger.YouShouldBeWorking);
        Assert.Equal(MentalState.Working, _sm.State);

        _sm.Fire(MentalTrigger.YouShouldBeWorking);
        Assert.Equal(MentalState.WorkingHarder, _sm.State); // WORKS!
    }
}
```
It still seems to work if _Fire()_ is called instead of _FireAsync()_, but I'm not sure I understand why that would be so? The synchronicity of a trigger or guard should not change the behaviour of the state machine, should it?
What am I doing wrong?

[...]

PR #626, opened 2025-08-21
Fix for transition precedence no longer given to substates

mclift:
This PR addresses issue #624, in which a breaking change in v5.18.0 caused asynchronous substate transitions to be ignored when a viable parent state transition is configured. These changes are intended to restore the transition precedence from prior releases.

* Fixed the trigger handler selection logic in `StateRepresentation.Async.cs` so that substate (local) transitions are preferred over superstate transitions in async scenarios. This ensures correct transition precedence when both substate and superstate can handle the same trigger.

* Added new tests in `TransitionFixture.cs` to verify that substate transitions correctly override superstate transitions for both synchronous and asynchronous trigger firing.

* Improved test naming and clarity in `TransitionFixture.cs` for reentry and transition tests, making the intent of each test clearer.

### DIFF
--- a/test/Stateless.Tests/TransitionFixture.cs
+++ b/test/Stateless.Tests/TransitionFixture.cs
@@ -1,25 +1,26 @@
-﻿using Xunit;
+﻿using System.Threading.Tasks;
+using Xunit;
 
 namespace Stateless.Tests
 {
     public class TransitionFixture
     {
         [Fact]
-        public void IdentityTransitionIsNotChange()
+        public void IsReentry_ShouldBeTrue_WhenSourceAndDestinationAreEqual()
         {
             StateMachine<int, int>.Transition t = new StateMachine<int, int>.Transition(1, 1, 0);
             Assert.True(t.IsReentry);
         }
 
         [Fact]
-        public void TransitioningTransitionIsChange()
+        public void IsReentry_ShouldBeFalse_WhenSourceAndDestinationDiffer()
         {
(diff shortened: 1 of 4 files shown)


test/Stateless.Tests/TransitionFixture.cs
```diff
@@ -1,25 +1,26 @@
-﻿using Xunit;
+﻿using System.Threading.Tasks;
+using Xunit;
 
 namespace Stateless.Tests
 {
     public class TransitionFixture
```
RedOrc: Very minor @mclift, but can we be explicit about the class name? I think changing this from **TransitionFixture** to **TransitionFixtureTest** is better. Thoughts?

mclift: I think I'm right in saying that the "Fixture" nomenclature was simply in common use when this library was first written and that just "Tests" would be more idiomatic now. Given there's a mixture of both in the project, perhaps renaming them all would be the right thing to do?

RedOrc: @mclift Agreed. If you like, I can take care of them after my other [PR](https://github.com/dotnet-state-machine/stateless/pull/625) is merged.

mclift: @RedOrc Sounds good. I've done this one in the meantime.

RedOrc (approved):

Issue #629, opened 2025-08-27
Substate transition 'hides' superstate transition in v5.19.0


MattVane:
Hi again! After upgrading to v5.19.0 I noticed that there is a similar issue persisting.

It appears now that if a) superstate and substate both handle the same trigger and b) guard condition on the substate transition that evaluates to false, then transition on the superstate is not invoked (and in fact the call to _FireAsync()_ fails). This does not happen with the synchronous _Fire()_ method. 

```csharp
public class StatelessTest2
{
    private readonly StateMachine<MentalState, MentalTrigger> _sm = new(MentalState.Working);
    private bool _hadCaffeine = true;

    public StatelessTest2()
    {
        _sm
            .Configure(MentalState.Awake)
            .PermitIf(MentalTrigger.Tired, MentalState.Napping);
        // Overrides the superstate transition
        _sm
            .Configure(MentalState.Working)
            .SubstateOf(MentalState.Awake)
            .PermitIf(MentalTrigger.Tired, MentalState.Grouchy, () => !_hadCaffeine);
    }

    private enum MentalState
    {
        Awake,
        Working,
        Grouchy,
        Napping
    }

    private enum MentalTrigger
    {
        Tired,
    }

    [Fact]
    public void GivenTriggerHandledOnSuperStateAndSubState_AndSubstateTransitionGuardBlocked_WhenTriggerFired_ThenShouldUseSuperstateTransition()
    {
        _sm.Fire(MentalTrigger.Tired);
        Assert.Equal(MentalState.Napping, _sm.State); // WORKS!
    }

    [Fact]
    public async Task GivenTriggerHandledOnSuperStateAndSubState_AndSubstateTransitionGuardBlocked_WhenTriggerFiredAsync_ThenShouldUseSuperstateTransition()
    {
        await _sm.FireAsync(MentalTrigger.Tired); // FAILS! No valid leaving transitions...
        Assert.Equal(MentalState.Napping, _sm.State); 
    }

    [Fact]
    public async Task GivenTriggerHandledOnSuperStateAndSubState_AndSubstateTransitionGuardIsOpen_WhenTriggerFiredAsync_ThenShouldUseSubstateTransition()
    {
        _hadCaffeine = false;
        await _sm.FireAsync(MentalTrigger.Tired);
        Assert.Equal(MentalState.Grouchy, _sm.State); 
    }
}
```

mclift:
[...]

PR #631, opened 2025-09-08
Fix for ignored async superstate transitions when child state has unmet guard

mclift:
This PR is for issue #629 in which a parent state transition is ignored when:

 - the child state has a transition configured with the same trigger;
 - the child state's transition has an unmet guard condition;
 - the parent state's transition does not have unmet guard conditions;
 - the child state is the current state;
 - the trigger is fired using `FireAsync`.

When the synchronous call to `Fire` is used instead, it selects the nearest ancestor that has a transition with no unmet guard conditions.

This PR also includes some housekeeping changes:

 - Removed unused `#if TASKS` directives.
 - Changed test method signatures in `AsyncFiringModesFixture` from `void` to `async Task`.
 - Split `TransitionTests` into sync and async test classes.

### DIFF
--- a/src/Stateless/OnTransitionedEvent.cs
+++ b/src/Stateless/OnTransitionedEvent.cs
@@ -21,7 +21,6 @@ namespace Stateless
                 _onTransitioned?.Invoke(transition);
             }
 
-#if TASKS
             public async Task InvokeAsync(Transition transition, bool retainSynchronizationContext)
             {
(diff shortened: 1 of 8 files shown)


mclift:
@MattVane @RedOrc review comments welcomed!

RedOrc (approved):
Looks good to me :-)

mclift:
I'll get this merged to dev so a pre-release package gets generated.

PR #639, opened 2025-12-19
Implemented FiringMode.Serial with standard locks

### LINKED ISSUE #527: Introduce FiringMode.Serial

tedchirvasiu:
This PR proposes an implementation for `FiringMode.Serial` as discussed in #527 

- `FiringMode.Serial` (`InternalFireSerial[Async]`) is pretty much a replica of `FiringMode.Queued` (`InternalFireQueued[Async]`) but it employs a classic short-lived lock to enqueue and dequeue elements, as discussed here: https://github.com/dotnet-state-machine/stateless/issues/527#issuecomment-3671468304

- The lock does not span the actual processing of the events, so it should not deadlock in any situation (even when Fire is called recursively from OnEntry/OnEntryAsync etc.)

- The lock should ensure only one thread can be the main "driving" thread of the state machine at a given time, thus ensuring triggers are processed sequentially just like in the normal queued mode (even when Fire is invoked from parallel threads).

```csharp
            lock (_serialModeLock) 
            {
                // Add trigger to queue
                _eventQueue.Enqueue(new QueuedTrigger { Trigger = trigger, Args = args });

                // If a trigger is already being handled then the trigger will be queued (FIFO) and processed later.
                if (_firing)
                    return;

                _firing = true;
            }
```

- There is one caveat, which is if the inner `InternalFireOne[Async]` throws whilst the state machine is being bombarded with events. Since the lock does not span the call to `InternalFireOne`, between the moment the `while` loop is broken and the `_firing` variable is set to false, other events might have been queued by other threads and those threads may not pick the role of the "driving thread" as `_firing` was still true.
So in this (probably rare) situation, the eventQueue might still contain unprocessed triggers, but no thread to drive the processing. Here I see two options:

1. Either the user has to trigger another `Fire` call (however since there is not parameter-less `Fire` overload, the user would be forced to actually implement a no-op trigger)
2. Or the eventQueue should be cleared in order to prevent leaving the state machine in an inconsistent state.

I personally think option 2 would be the sanest in this parallel scenario if you can afford to lose triggers. For this scenario I implemented the `DropUnprocessedEventsOnErrorInSerialMode` prop which by default is false (so no lost triggers, but the SM might need a kick to resume execution).

- Some additional tests may be needed, for instance for the new `DropUnprocessedEventsOnErrorInSerialMode` parameter.

Resolves #527

### DIFF
--- a/README.md
+++ b/README.md
@@ -360,11 +360,11 @@ Setting this is vital within a Microsoft Orleans Grain for example, which requir
 
 ### Thread safety
 By default, Stateless is **NOT** thread-safe.
-`FiringMode.Sequential` ensures thread-safety for Fire(), however reading the State Machine's state from multiple threads may still be unsafe and require aditional locks.
+`FiringMode.Serial` ensures thread-safety for Fire(), however reading the State Machine's state from multiple threads may still be unsafe and require aditional locks.
 
 Stateless processes triggers sequentially, and as a result there can only be one thread "driving" the processing at a time.
 
-In `FiringMode.Sequential`, if the main processing thread throws an error, unprocessed triggers should be removed from the queue in order to ensure consistency. Otherwise the event queue may still hold unprocessed triggers which would require another Fire() call to resume processing.  
+In `FiringMode.Serial`, if the main processing thread throws an error, unprocessed triggers should be removed from the queue in order to ensure consistency. Otherwise the event queue may still hold unprocessed triggers which would require another Fire() call to resume processing.  
 Set `DropUnprocessedEventsOnErrorInSerialMode` to true if you need consistent behaviour.  
 Set `DropUnprocessedEventsOnErrorInSerialMode` to false if you don't want triggers to be dropped (default).
 
--- a/README.md
+++ b/README.md
@@ -358,6 +358,16 @@ var stateMachine = new StateMachine<State, Trigger>(initialState)
 
 Setting this is vital within a Microsoft Orleans Grain for example, which requires the `SynchronizationContext` in order to make calls to other Grains.
 
+### Thread safety
+By default, Stateless is **NOT** thread-safe.
+`FiringMode.Sequential` ensures thread-safety for Fire(), however reading the State Machine's state from multiple threads may still be unsafe and require aditional locks.
+
+Stateless processes triggers sequentially, and as a result there can only be one thread "driving" the processing at a time.
+
+In `FiringMode.Sequential`, if the main processing thread throws an error, unprocessed triggers should be removed from the queue in order to ensure consistency. Otherwise the event queue may still hold unprocessed triggers which would require another Fire() call to resume processing.  
+Set `DropUnprocessedEventsOnErrorInSerialMode` to true if you need consistent behaviour.  
+Set `DropUnprocessedEventsOnErrorInSerialMode` to false if you don't want triggers to be dropped (default).
+
 ## Building
 
 Stateless runs on .NET runtime version 4+ and practically all modern .NET platforms by targeting .NET Framework 4.6.2, .NET Standard 2.0, and .NET 8.0, 9.0 and 10.0. Visual Studio 2017 or later is required to build the solution.
--- a/src/Stateless/StateMachine.Async.cs
+++ b/src/Stateless/StateMachine.Async.cs
@@ -162,9 +162,71 @@ namespace Stateless
                 case FiringMode.Queued:
                     await InternalFireQueuedAsync(trigger, args);
                     break;
+                case FiringMode.Serial:
+                    await InternalFireSerialAsync(trigger, args);
+                    break;
                 default:
                     // If something is completely messed up we let the user know ;-)
                     throw new InvalidOperationException("The firing mode has not been configured!");
+            }
+        }
+
+        /// <summary>
+        /// Queue events and then fire in order.
+        /// If only one event is queued, this behaves identically to the non-queued version.
+        /// </summary>
+        /// <param name="trigger">  The trigger. </param>
+        /// <param name="args">     A variable-length parameters list containing arguments. </param>
+        async Task InternalFireSerialAsync(TTrigger trigger, params object[] args) {
+
+            lock (_serialModeLock) 
+            {
+                // Add trigger to queue
+                _eventQueue.Enqueue(new QueuedTrigger { Trigger = trigger, Args = args });
+
+                // If a trigger is already being handled then the trigger will be queued (FIFO) and processed later.
+                if (_firing)
+                    return;
+
+                _firing = true;
+            }
+
+            try 
+            {
+
+                // Empty queue for triggers
+                while (true)
+                {
+
+                    QueuedTrigger queuedEvent;
+
+                    lock (_serialModeLock)
+                    {
+
+                        if (_eventQueue.Count == 0) 
+                        {
+                            _firing = false;
+                            break;
+                        }
+
+                        queuedEvent = _eventQueue.Dequeue();
+                    }
+
+                    await InternalFireOneAsync(queuedEvent.Trigger, queuedEvent.Args).ConfigureAwait(RetainSynchronizationContext);
+                }
+            } 
+            catch
+            {
+
+                lock (_serialModeLock)
+                {
+                    if (DropUnprocessedEventsOnErrorInSerialMode)
+                        _eventQueue.Clear();
+
+                    _firing = false;
+                }
+
+                throw;
             }
         }
 
--- a/src/Stateless/StateMachine.cs
+++ b/src/Stateless/StateMachine.cs
@@ -14,7 +14,9 @@ namespace Stateless
         /// <summary> Use immediate mode when the queuing of trigger events are not needed. Care must be taken when using this mode, as there is no run-to-completion guaranteed.</summary>
         Immediate,
         /// <summary> Use the queued <c>Fire</c>ing mode when run-to-completion is required. This is the recommended mode.</summary>
-        Queued
+        Queued,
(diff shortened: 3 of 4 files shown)


src/Stateless/StateMachine.Async.cs
```diff
@@ -162,12 +162,74 @@ async Task InternalFireAsync(TTrigger trigger, params object[] args)
...
 
+        /// <summary>
+        /// Queue events and then fire in order.
+        /// If only one event is queued, this behaves identically to the non-queued version.
+        /// </summary>
+        /// <param name="trigger">  The trigger. </param>
+        /// <param name="args">     A variable-length parameters list containing arguments. </param>
+        async Task InternalFireSerialAsync(TTrigger trigger, params object[] args) {
+
+            lock (_serialModeLock)
```
gao-artur, line 182: Using a regular lock in an async method isn't a best practice as it blocks the thread during the wait for the lock. Consider using `SemaphoreSlim(1,1)` that supports both sync and async waiting instead.

README.md
```diff
@@ -358,6 +358,16 @@ var stateMachine = new StateMachine<State, Trigger>(initialState)
 
 Setting this is vital within a Microsoft Orleans Grain for example, which requires the `SynchronizationContext` in order to make calls to other Grains.
 
+### Thread safety
+By default, Stateless is **NOT** thread-safe.
+`FiringMode.Sequential` ensures thread-safety for Fire(), however reading the State Machine's state from multiple threads may still be unsafe and require aditional locks.
+
+Stateless processes triggers sequentially, and as a result there can only be one thread "driving" the processing at a time.
```
gao-artur, line 365: This is very problematic behavior, especially in an environment with high load on SM. Consider that the first caller starts processing the event, and then 100 more callers enqueue additional events. The first caller will continue processing all these events, which may take a considerable amount of time and, in extreme cases, may be ongoing indefinitely, blocking the first caller forever.
A better solution, in my opinion, would be to start an internal worker thread that pulls tasks to execute from an internal queue (a dedicated queue for `Serial` mode, not the `_eventQueue`). Fire methods will add their events to this queue.
However, this introduces another complexity: do callers expect the `Fire(Async)` method to return immediately after queuing their tasks, or only after their task has been processed? I think both scenarios have valid usages, and both should be supported by the configuration flag. The blocking scenario can be achieved by enqueuing a complex object that, in addition to the trigger, will also hold `TaskCompletionSource` instance. The `InternalFireSerial(Async)` will wait on its `Task`, and the worker thread will `SetResult`/`SetException/SetCancelled` when done/failed/canelled. The sync `InternalFireSerial` method can call `Wait` on the `TaskCompletionSource.Task` without a risk of sync-over-async problems, because this `Task` is not async, so deadlock is not possible even in environments with `SynchronizationContext`.
The blocking scenario introduces another problem: the waiting may need to be cancelled, so `Fire` methods reqiure support for `CancellationToken` (both sync and async).
And the final one: need a way to stop the worker thread gracefully, so the state machine should now implement `I(Async)Disposable`.

README.md
```diff
@@ -358,6 +358,16 @@ var stateMachine = new StateMachine<State, Trigger>(initialState)
 
 Setting this is vital within a Microsoft Orleans Grain for example, which requires the `SynchronizationContext` in order to make calls to other Grains.
 
+### Thread safety
+By default, Stateless is **NOT** thread-safe.
+`FiringMode.Sequential` ensures thread-safety for Fire(), however reading the State Machine's state from multiple threads may still be unsafe and require aditional locks.
```
gao-artur: You call it `Serial` everywhere, but `Sequential` in documentation.

src/Stateless/StateMachine.cs
```diff
@@ -40,9 +42,20 @@ private class QueuedTrigger
...
 
+        /// <summary>
+        /// If the main processing thread throws an error, unprocessed triggers should be removed from
+        /// the queue in order to ensure consistency. Otherwise the event queue
+        /// may still hold unprocessed triggers which would require another
+        /// Fire() call to resume processing.
+        /// Set this to true if you need consistent behaviour.
+        /// Set this to false if you don't want triggers to be dropped.
+        /// </summary>
+        public bool DropUnprocessedEventsOnErrorInSerialMode { get; set; } = false;
```
gao-artur, line 57: I think this and other configurations should be immutable. They should be set on `StateMachine` creation, and their change should not be allowed after that.

src/Stateless/StateMachine.Async.cs
tedchirvasiu: Please take a look at [SemaphoreSlim's source code](https://github.com/dotnet/dotnet/blob/b0f34d51fccc69fd334253924abd8d6853fad7aa/src/runtime/src/libraries/System.Private.CoreLib/src/System/Threading/SemaphoreSlim.cs#L683).

**It also uses a regular lock** under the hood. In addition to that, it may cause a thread switch. Considering all the overhead, I really don't think it's best practice or more performant to use a SemahporeSlim for a very short lived lock such as flipping a boolean and pushing an element to a Queue, especially since no deadlocking can occur.

README.md
tedchirvasiu: Thank you, fixed.

src/Stateless/StateMachine.cs
tedchirvasiu: Thought about it too, but that would require adding yet another constructor or another parameter to the existing ones.

Is it worth it?

Because I think nobody would be crazy enough to change this while the state machine is running, right?

gao-artur: It's not about crazy users. It's about preventing mistakes by less experienced developers. Is it possible to use `init` instead of `set` with supported TFM's?

README.md
tedchirvasiu: > This is very problematic behavior, especially in an environment with high load on SM. Consider that the first caller starts processing the event, and then 100 more callers enqueue additional events. The first caller will continue processing all these events, which may take a considerable amount of time and, in extreme cases, may be ongoing indefinitely, blocking the first caller forever.

In the Serial mode, technically no caller is required to await on FireAsync. So you can fire and forget if you are worried about having the initial task blocked, but just as in the case of having a separate worker thread for processing, you lose the ability to try-catch errors (unless you use Task.Run and add the error handling in there too)

Having a dedicated worker thread for processing events would be great, but I think it would pretty much require a totally different architecture than the one that exists. Think about exception handling. If you queue all work on a Task completely separate from your caller Tasks, exception handling would no longer work with a simple try-catch surrounding the call to FireAsync. You would need a separate OnError event or as you suggested, a dedicated TaskCompletionSource for each queued InternalFireOneAsync.

**Also, another question is what happens when processing an event throws an error? Do you keep going with the next event? Do you halt and require a separate function call to resume?**

In the current FiringMode.Queued implementation the execution halts and it would only resume on the next FireAsync call (since events don't get dropped).

> However, this introduces another complexity: do callers expect the Fire(Async) method to return immediately after queuing their tasks, or only after their task has been processed?

In the existing Queued mode, callers can't expect FireAsync to return only after THEIR task has been processed because it returns immediately after queueing if `_firing` is true. So in Serial mode they could in theory always expect to have the task return as soon as the item has been queued and it would still be consistent.

tedchirvasiu: The intention in this PR was to propose the simplest form of thread safety without entirely changing the library's design around supporting multi-threading.

Reading the maintainers' comments on the original issue and on other PRs regarding this issue, I got the feeling they are pretty reluctant to implementing Serial mode in fear of overcomplicating the library.

I support the idea of having a dedicated worker thread, but I personally can't think of a simple, elegant and safe enough implementation.

src/Stateless/StateMachine.cs
tedchirvasiu: Hmm, I think it might work with preprocessor symbols, but I haven't done it before on a project supporting all .Nets. Do you have a working snippet by any chance?

> It's not about crazy users. It's about preventing mistakes by less experienced developers

But I think even if someone were to flip this while the SM is running, nothing would blow up.

README.md
tedchirvasiu: @gao-artur This should be fixed in https://github.com/dotnet-state-machine/stateless/pull/640

